PR #71, opened 2025-07-20
[EXPERIMENT] Use SDL2 For 3DS

NateXS:
To-Do list:
- [x] Get it to compile
- [x] Render images to the screen
- [ ] Input support
- [x] Audio support

I spent a couple hours trying to see if using SDL2 for rendering was a viable alternative to Citro2D. A lot of that time was spent trying to compile SDL2 for 3DS since its not in dkp-pacman (only SDL1 is) and there was little info on how to compile,,,,

But I did get it working. It currently uses code from the `sdl` folder and no code from the `3ds/` folder. The results...? Well, It's fine if you're not rendering images, but as soon as an image needs to render, the FPS tanks.
 
https://github.com/user-attachments/assets/aa36631c-0794-4777-9d9b-65cc1bf41690

(Citro2D rendering on the left, SDL2 rendering on the right)

It's _very_ likely the reason is because SDL2 for 3DS has to use software rendering, instead of GPU rendering like Citro does. There's also other issues with it like longer loading times, and the game going black when going to the home menu.

I do want to see though if SDL2 is a good option for implementing audio, which is why I'm leaving this here.


gradylink:
Oh, nice! Once my Wii U PR gets merged, the same controller code should work for the 3DS.

gradylink:
Just noticed. This PR modifies the `initApp` (and `Render::Init`) to return a boolean for error handling. My Wii U PR modifies these functions to return integers for error handling. Would you like me to update the Wii U PR to use booleans instead of integers

NateXS:
> Just noticed. This PR modifies the `initApp` (and `Render::Init`) to return a boolean for error handling. My Wii U PR modifies these functions to return integers for error handling. Would you like me to update the Wii U PR to use booleans instead of integers

That would be great yeah

source/main.cpp
```diff
@@ -12,12 +12,17 @@ static void exitApp() {
     Render::deInit();
 }
 
-static void initApp() {
-    Render::Init();
+static bool initApp() {
+    if (!Render::Init())
```
gradylink: Couldn't this just be:

```cpp
return Render::Init();
```

gradylink: ? It went up 2 lines instead of down two lines. And I clicked make a comment, not start a review.

NateXS: > Couldn't this just be:
> 
> ```c++
> return Render::Init();
> ```

Yeah you're right, I'll fix that real quick

NateXS:
closing this as I got everything I wanted out of it 😁

PR #80, opened 2025-07-20
feat: init 3ds nightly build

gradylink:
Inspired by https://github.com/NateXS/Scratch-3DS/discussions/34#discussioncomment-13808967.

This will create a nightly build available to logged-in users in the actions tab. It shouldn't require any additional configuration of environment variables.

### DIFF
--- a/.github/workflows/nightly-3ds.yml
+++ b/.github/workflows/nightly-3ds.yml
@@ -1,0 +1,41 @@
+on:
+  push:
+    paths:
+      - Makefile_3ds
+      - source/scratch/**
+      - source/3ds/**
+      - gfx/**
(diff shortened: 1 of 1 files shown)


NateXS:
Would using `path: Scratch.3dsx` instead of `path: Scratch-3DS.3dsx` make more sense, since Scratch.3dsx is what I use for the official builds?

gradylink:
Well, that's the path that the Makefile spits out. It should be possible to use `mv` to convert it to `Scratch.3dsx` before publishing it, if you'd like me to do that.

NateXS:
actually Scratch-3DS is fine, thanks!

PR #83, opened 2025-07-21
feat: use devkitarm container for nightly action

gradylink:
This makes it work. When you merge this, I would recommend doing it squashed as most of these commits were unsuccessful in fixing the problem.


NateXS:
merged, thanks!

PR #125, opened 2025-07-23
fix: prioritize costume names over indexes

### LINKED ISSUE #118: [BUG] Costumes prioritize number type over string
**Describe the bug**
When using the "switch costume to ()" block, if a number is provided as the input, it prioritizes using the input as a number over as a string. This is inconsistent with Scratch, which prioritizes using the input as a string (in the event that a costume exists with a number as a name).

**Screenshots**
To the left of the stage in Turbowarp, the number 16 can be seen, displayed using sprites with costumes named "1" and "6":
Image

To the left of the stage in Scratch 3DS, "!)" is instead seen, as the sprite's first ("!") and sixth costumes (")") are used instead:
Image

**Device**
New Nintendo 3DS

gradylink:
Fixes #118

### DIFF
--- a/source/scratch/blocks/looks.cpp
+++ b/source/scratch/blocks/looks.cpp
@@ -21,7 +21,18 @@ BlockResult LooksBlocks::switchCostumeTo(Block &block, Sprite *sprite, Block **w
         }
     }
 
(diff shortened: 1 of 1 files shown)


NateXS:
merged, thanks!

PR #248, opened 2025-08-15
Initial main menu update and custom controls

NateXS:
Been working on this one for a bit. The look of the main menu will be pretty bare-bones for this PR as I just wanted to get the functionality working first. UI programming is hard.......
(Also if anyone can make better button icons please do because the current ones don't look great 😭)

https://github.com/user-attachments/assets/56a7b33c-e56f-488d-b143-94df6a540abd

PC version as well to get a reference for how it looks on other platforms:

https://github.com/user-attachments/assets/2cf5e9a2-dcaf-42a1-80df-9ddd9838f77f

### Custom controls
- Configurations get saved and loaded from `scratch-everywhere/'project_name'.json`
- - Wii U is `sd:/wiiu/scratch-wiiu/'project_name'.json`
- Works by loading the project and looking for every `Key pressed` block

### Other
- `Stop All` block takes you back to the main menu now instead of homebrew menu
- [3DS] new loading screen

Once this PR is merged is probably when I'll change the repo name. It should be good to merge, I just need to test it out on every platform first


gradylink:
> UI programming is hard.......

I hate UI dev... Maybe we could use a layout library like Clay?

EDIT:

> (Also if anyone can make better button icons please do because the current ones don't look great 😭)

I'll see what I can do. I'll be busy for the rest of today though, so feel free to merge this, then I can update them later.

NateXS:
ok tested on 3DS, Wii, Wii U, and emulated Gamecube. Works! I don't have a switch emulator set up, and there's a 50/50 chance custom controls won't work since I think the Switch also uses a different folder like the Wii U. May have to fix that. Anyways I'm gonna merge!

PR #276, opened 2025-08-21
[WIP] feat: init OG Xbox support

gradylink:
# Checklist

- [ ] Make it build (Miniz issues)
- [ ] Make it not flood the source directory with build stuff
- [ ] Input
- [ ] Cloud Variables
- [ ] Rendering
- [ ] Loading Projects
- [ ] Dockerfile


gradylink:
I think the only way to fix the build errors is to migrate to zlib on SDL2 builds because SDL2_image relies on it anyway.

EDIT: never mind zlib can't do what we need to do, maybe I can get libzip or minizip to work.

EDIT 2: i've started rewriting Scratch Everywhere! to use Minizip instead of Miniz, and it's been pretty annoying...

gradylink:
This is too annoying to work on, so i'm gonna close it.

PR #287, opened 2025-08-23
W.I.P. Arch Build

ErrorOliver2:
haven't tested it yet though, just changed apt-get with pacman which is the default package manager on arch and added base-devel (not sure if this helps or not?)

might test this later


gradylink:
What? It already builds on Arch? That's mostly just for nightly builds.

ErrorOliver2:
oh, i guess im being dumb then sorry, i've not had my sleep

PR #290, opened 2025-08-24
Update README for Vita and clean up grammar/spelling/readability

Dogo6647:
This PR will:
☑ Add instructions for setting up the Vita in both installation and building
☑ Fix grammar/spelling mistakes, redundant or misplaced punctuation, ensure ***correct usage of your and you're***, and make the file a bit easier to read and digest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
   <img src="gfx/logo.png" alt="Scratch Everywhere! logo" />
 </p>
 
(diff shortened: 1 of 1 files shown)


Dogo6647:
I kept most of the readme's format as-is, but I'm telling y'all; We might need to set up a GitHub page with the docs or something at some point because the readme's getting way too long.

gradylink:
> I kept most of the readme's format as-is, but I'm telling y'all; We might need to set up a GitHub page with the docs or something at some point because the readme's getting way too long.

I agree. I might set up a Vitepress site.

PR #295, opened 2025-08-25
Add list support to monitor display

Br0tcraft:
The changes are not very big, but with this it should also be possible to display lists.
A list is treated as a single string, using \n as a separator so that each element appears on a new line.


NateXS:
Looks good to me. Merging!

PR #297, opened 2025-08-25
Add loudness variable

Br0tcraft:
I added support for the loudness variable block.
It now returns a value from 0 to 100. If the device isn’t a 3DS or the mic can’t be initialized, it returns -1.

There’s a MIC_SENSITIVITY variable in mic.cpp (currently 30.0). This might still need some tweaking to get values closer to Scratch.

I’m not sure how the microphone behaves on a real 3DS, but in Citra it might react a bit differently than on real hardware (could be more or less sensitive).

### DIFF
--- a/source/scratch/mic.hpp
+++ b/source/scratch/mic.hpp
@@ -1,0 +1,15 @@
+#pragma once
+#include <math.h>
+
+#ifdef __3DS__
+#include <3ds.h>
+#include <malloc.h>
+#endif
+
+extern u8* micbuf;
+extern u32 micbuf_datasize;
+extern bool micInitialized;
+
+int getMicLevel();
+bool initMic();
+void exitMic();
--- a/source/3ds/render.cpp
+++ b/source/3ds/render.cpp
@@ -6,6 +6,7 @@
 #include "../scratch/render.hpp"
 #include "../scratch/text.hpp"
 #include "../scratch/unzip.hpp"
+#include "mic.hpp"
 #include "image.hpp"
 #include "interpret.hpp"
 #ifdef ENABLE_AUDIO
@@ -486,6 +487,7 @@ void Render::deInit() {
 #ifdef ENABLE_AUDIO
     SDL_Quit();
 #endif
+    exitMic();
     romfsExit();
     gfxExit();
 }
--- a/source/scratch/blockExecutor.cpp
+++ b/source/scratch/blockExecutor.cpp
@@ -163,6 +163,7 @@ void BlockExecutor::registerHandlers() {
     valueHandlers["sensing_touchingobjectmenu"] = SensingBlocks::touchingObject; // Menu variant
     valueHandlers["sensing_mousedown"] = SensingBlocks::mouseDown;
     valueHandlers["sensing_username"] = SensingBlocks::username;
+    valueHandlers["sensing_loudness"] = SensingBlocks::loudness;
 
     // procedures / arguments
     handlers["procedures_call"] = ProcedureBlocks::call;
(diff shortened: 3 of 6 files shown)


gradylink:
I think we should use SDL2 for this on all platforms because we use SDL2 for audio out on 3DS, so I think we should use it for audio in too. Also, it should only be enabled with `ENABLE_AUDIO` because it needs that for SDL2 on 3DS.

Br0tcraft:
I could try to make that work, but I thought SDL2 didn’t support microphone input on the 3DS. Are there actually ways to do it?

gradylink:
> I could try to make that work, but I thought SDL2 didn’t support microphone input on the 3DS. Are there actually ways to do it?

Oh, yeah, I just checked, it doesn't:

```c
static int N3DSAUDIO_CaptureFromDevice(_THIS, void *buffer, int buflen)
{
    /* Delay to make this sort of simulate real audio input. */
    SDL_Delay((this->spec.samples * 1000) / this->spec.freq);

    /* always return a full buffer of silence. */
    SDL_memset(buffer, this->spec.silence, buflen);
    return buflen;
}
```

It would still be nice to add SDL2 support for the other platforms, so I would recommend moving `source/scratch/mic.cpp` to `source/3ds/audio/mic.cpp` and then create an equivalent file for SDL2.

Br0tcraft:
I guess for the SDL ports we’d probably just have to use static values (like always -1), since I don’t think real mic input is supported on the other platforms. Or are there actually ways (through SDL) ?

NateXS:
> I guess for the SDL ports we’d probably just have to use static values (like always -1), since I don’t think real mic input is supported on the other platforms. Or are there actually ways (through SDL) ?

yup, SDL does support mic input

source/scratch/mic.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+#include <math.h>
+
+#ifdef __3DS__
+#include <3ds.h>
+#include <malloc.h>
+#endif
+
+extern u8* micbuf;
+extern u32 micbuf_datasize;
```
NateXS:
```suggestion
extern u_int8_t* micbuf;
extern u_int32_t micbuf_datasize;
```
The variable types should be the full name here since the shortened types are a 3ds.h thing

gradylink: The rest of the project seems to not of the underscore between the `u` and the `i` so I think we should use:

```
extern uint8_t* micbuf;
extern uint32_t micbuf_datasize;
```

Instead.

gradylink:
> > I guess for the SDL ports we’d probably just have to use static values (like always -1), since I don’t think real mic input is supported on the other platforms. Or are there actually ways (through SDL) ?
> 
> yup, SDL does support mic input

Hmm, so SDL2 itself supports mic input (so we should still support it) but the only SDL2 device we support that actually implements it is the Vita. The rest just tell SDL2 that they don't have a microphone (which does make sense for all of them except Wii U, I might make a PR.)

Br0tcraft:
I tried to write a script to make it work with SDL as well.
For some reason, I can only compile for the 3DS, compiling for other platforms fails even without my SDL code.
I wasn’t able to test if it actually works.
Should I still add it here, or could someone else take care of it?

gradylink:
> I tried to write a script to make it work with SDL as well. For some reason, I can only compile for the 3DS, compiling for other platforms fails even without my SDL code. I wasn’t able to test if it actually works. Should I still add it here, or could someone else take care of it?

It would be nice to add it.

Br0tcraft:
Okay, that's done. I don't know if that works, but I'd also have to call exitMic() somewhere (for sdl).

Br0tcraft:
@NateXS 
Where exactly can I call ExitMic() (SDL)?
After that the pull request would also be ready to merge.

NateXS:
> @NateXS Where exactly can I call ExitMic() (SDL)? After that the pull request would also be ready to merge.

you can do it in `sdl/render.cpp` `Render::deInit()`

PR #322, opened 2025-09-02
Implement writeToFile

turbotoad12:
I added the functionality to the writeToFile method and that the log functions will write to the log file

> PS it says theres 5 commits but theres only one. ill update my repo


turbotoad12:
Let me fix this

PR #336, opened 2025-09-15
Change Username in Settings

Br0tcraft:
The settings button is now displayed next to the block icon. 
In the settings, you can enable "Custom Username" and assign a custom name. 
This name will also be used for cloud projects, but the file where the cloud name is stored will not be overwritten. Therefore, if you disable Costume Username again, the old cloud name will be used.

Currently, a maximum of 9 characters and only A-Z, a-z, 0-9, and "_" are allowed. 

I've only tried it on the 3DS so far.


source/scratch/interpret.cpp
```diff
@@ -117,6 +120,32 @@ void initMist() {
...
+                    if (std::isalnum(static_cast<unsigned char>(c)) || c == '_') {
+                        hasNonSpace = true;
+                    } else if (!std::isspace(static_cast<unsigned char>(c))) {
+                        break;
+                    }
+                }
+                if(hasNonSpace) customUsername = j["Username"].get<std::string>();
+                else customUsername = "Player";
+            }
+        }
```
gradylink: This doesn't look properly formatted to me, can you run `clang-format` (on all files.)

Br0tcraft: @gradylink oh sorry, but should be done now

gradylink (approved):

PR #337, opened 2025-09-17
[Feature] Allow splitting large projects into smaller ones

Br0tcraft:
Some users have run into issues when trying to launch Scratch projects in Scratch Everywhere, mostly because their projects were simply too big.
To make this a bit easier, I thought about adding a way to switch .sb3 files directly inside a running project.

Of course, this doesn’t magically make oversized Scratch projects work on SE!,  but it could give developers the option to split a large project into multiple smaller ones and then switch between them internally.

Right now, this is just a prototype. It’s basically one of the “My Blocks” that I tweaked (similar to how the logging blocks were modified). It works like this: "open %s .sb3" with "arg0" as the parameter.

What’s still missing is some way to share data between those smaller projects, so this is definitely not finished yet.

Not sure if this is really useful as a proper feature, but I wanted to put the idea out there and hear your thoughts.

### DIFF
--- a/source/scratch/blockExecutor.cpp
+++ b/source/scratch/blockExecutor.cpp
@@ -11,8 +11,8 @@
 #include "interpret.hpp"
 #include "math.hpp"
 #include "os.hpp"
-#include "unzip.hpp"
 #include "sprite.hpp"
+#include "unzip.hpp"
 #include <algorithm>
 #include <chrono>
 #include <cstddef>
@@ -328,7 +328,6 @@ BlockResult BlockExecutor::runCustomBlock(Sprite *sprite, Block &block, Block *c
     if (block.customBlockId == "\u200B\u200Blog\u200B\u200B %s") Log::log("[PROJECT] " + Scratch::getInputValue(block, "arg0", sprite).asString());
     if (block.customBlockId == "\u200B\u200Bwarn\u200B\u200B %s") Log::logWarning("[PROJECT] " + Scratch::getInputValue(block, "arg0", sprite).asString());
     if (block.customBlockId == "\u200B\u200Berror\u200B\u200B %s") Log::logError("[PROJECT] " + Scratch::getInputValue(block, "arg0", sprite).asString());
-    Log::log(block.customBlockId);
     if (block.customBlockId == "​open %s .sb3") {
         Log::log("Open next Project with Block");
         Scratch::nextProject = true;
--- a/source/scratch/blockExecutor.cpp
+++ b/source/scratch/blockExecutor.cpp
@@ -399,13 +399,9 @@ std::vector<Block *> BlockExecutor::runAllBlocksByOpcode(std::string opcodeToFin
                 // runBlock(data,currentSprite);
                 blocksRun.push_back(&data);
                 executor.runBlock(data, currentSprite);
-                // Wouldn't a 'goto' be useful here?
-                // Or do I just not understand this code well enough?
-                // goto returnBlocksRun;
             }
         }
     }
-    // returnBlocksRun:
     return blocksRun;
 }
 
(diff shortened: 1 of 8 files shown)


gradylink:
I was thinking about doing this before, I was thinking maybe as a custom extension, but this works as well (at least until custom extensions are added.) I think one good way to do it, is another custom block that is like `make %s transfer`, which will make whatever variable name passed to it transfer between project changes. That's actually something not possible with custom extensions (they would need to store it in a temporary file.)

EDIT: it would be kinda cool to make the main menu a scratch project in the future tbh

source/scratch/blocks/procedure.cpp
```diff
@@ -23,6 +24,13 @@ Value ProcedureBlocks::booleanArgument(Block &block, Sprite *sprite) {
 }
 
 BlockResult ProcedureBlocks::call(Block &block, Sprite *sprite, bool *withoutScreenRefresh, bool fromRepeat) {
+    if (block.customBlockId == "​open %s .sb3") {
+        Log::log("Open next Project with Block");
+        Scratch::nextProject = true;
+        Unzip::filePath = Scratch::getInputValue(block, "arg0", sprite).asString() + ".sb3";
+        Scratch::shouldStop = true;
+        return BlockResult::RETURN;
+    }
```
gradylink: This isn't actually the correct place for this, it should be at https://github.com/ScratchEverywhere/ScratchEverywhere/blob/021dab94c64966500e79c8cfe22548853919cb2c/source/scratch/blockExecutor.cpp#L327-L329 and wrapped in the `200B` (zero width space, Unicode character) like with the Scratch Addons blocks. I can generate a Scratch project with the special `200B` version of the block if you want.

source/scratch/blockExecutor.cpp
```diff
@@ -380,9 +380,13 @@ std::vector<Block *> BlockExecutor::runAllBlocksByOpcode(std::string opcodeToFin
                 // runBlock(data,currentSprite);
                 blocksRun.push_back(&data);
                 executor.runBlock(data, currentSprite);
+                // Wouldn't a 'goto' be useful here?
+                // Or do I just not understand this code well enough?
+                // goto returnBlocksRun;
```
gradylink: This runs **all** blocks with that opcode, so exiting the loop early (should be using `break` instead of `goto` anyway) would make it only run the first block (e.g. only run one flag block.)

Br0tcraft: oh the opcosde not the id. I guess I misread it, that makes more sense

And you're right "break" would fit there. I'm not that practiced with "Goto" yet, sorry.

source/scratch/blocks/procedure.cpp
Br0tcraft: Yes I had seen, but the function was a void function. And in order to then change a project in it, I would have had to change more. It wasn't meant to be merged at all. I just wanted to quickly build a prototype that works.

Br0tcraft: Yes, that would be great if you could do that with the 200b codes, although I don't think it should be too difficult. But something like the name of the block can be changed afterwards too.

Br0tcraft:
@gradylink  A scratch project as a launcher. That would be really a funny idea. Possibly if the necessary extensions are translated for reading files. (and unpacking files? If you then want to integrate the Unpacke Project Feature in Scratch form)

Br0tcraft:
I’ve restructured things a bit (at least I hope so) and also added two new blocks. One lets you open another project and pass data to it (open ____ .sb3 with data ____), and the other lets you retrieve that data from the last project (data last Project). I’m still not really sure what the best labels for these blocks should be, so that might need to be reworked later.

While testing, I noticed that using a path like MyGame/shop will open shop.sb3 from inside the MyGame folder (relative to the SE! root). That project doesn’t show up in the SE! project list, but it works just fine, which actually makes it pretty handy for building larger games without cluttering the list or accidentally starting the wrong project.

Br0tcraft:
@gradylink 
Is that ready to be merged or am I missing something?

gradylink:
> @gradylink Is that ready to be merged or am I missing something?

Could you just wrap the blocks in `\u200b` and send over a template project (so I can modify the blocks in there as well.) The I think we'll be good. You also need to update the README and you should probably say something like "this feature may be removed later and be replaced by custom extensions".

Br0tcraft:
Okay, should this be fine now?
Is the note in the readme okay, or is it perhaps too long/short?

source/scratch/blocks/procedure.cpp
```diff
@@ -9,7 +9,7 @@ Value ProcedureBlocks::stringNumber(Block &block, Sprite *sprite) {
     if (name == "Scratch Everywhere! platform") {
         return Value(OS::getPlatform());
     }
-    if (name == "data last Project") {
+    if (name == "received data") {
```
gradylink: This should be wrapped as well.

gradylink:
OK so um I think it's all fine, I'll go make a project containing the blocks (like the one for the SE! blocks) and put it on ScratchBox, and then you can link to it in the README.

EDIT: or i guess you should put it on ScratchBox, I'll DM you the updated project.

gradylink (approved):

PR #349, opened 2025-09-27
feat(3ds): Create a linker for Scratch projects

Br0tcraft:
Far from finished.
BUT you can already change the name and the author of a Cia (scratch-everywhere/Linker/Linker-v0.3.cia)

There are still things missing, such as the script to generate a project ID, the CIA being given the name of the Scratch project and generally the fine-tuning and error handling.


source/scratch/linker.cpp
gradylink, line 1: I think this should be handled similarly to `render.cpp` where there's one in each platform's directory and a shared header file in `source/scratch`.

Br0tcraft:
Everything works in Citra, but the CIA fails to install on real 3DS hardware.
My hypothesis: modifying a finished CIA breaks its verification data (TMD/ticket/content hashes, NCCH checksums), so the console refuses to install/run it. Emulators like Citra appear to ignore or not strictly enforce those signature/hash checks, which is why the modified CIA runs there. (My test earlier on real hardware were only individual bytes that were changed, but I used to test it mainly in Citra because I only tested it for fun to see if it would work) I thought the check would be more general whether the CIA makes sense and not whether something has been changed 

I tried to generate two CIAs that differ only in a small metadata field (e.g. title/ID) to compare differences. However, makerom produces non-identical binaries even with the same RSF/source: the generated CIAs contain different Title ID (0004000XXXXXXXXX) bytes and extra bytes in other fields.

Plan:

Extract TMD, ticket and NCCH (including RomFS/Exe) from both CIAs with ctrtool or GodMode9.

Diff the internal components to precisely locate differences (TMD, ticket, NCCH header, content files, signature blobs, etc.).

Inspect makerom / Project_CTR source paths that build/sign TMD/ticket/NCCH to see which fields are written and whether non-deterministic data (timestamps, random padding, non-deterministic signature data) is inserted.


If makerom produces nondeterministic output, that would explain why two byte-identical CIAs can’t be created. But I really think I'm doing something wrong when building the CITRA file.

BUT for now I'll take a break from this feature. Probably in the month or so when I have new vigor, ideas and approaches that could bring something I'll try again

PR #351, opened 2025-09-30
[Menu] Add Main Menu music

NateXS:
Pull request to add the award winning and critically acclaimed `mm_splash.ogg` and `mm_full.ogg` into the main menu 📈📈🔥

TO-DO List

- [x] Refactor audio so it can be used in the menu
- [x] Make the music smoothly transition between both tracks

- [x] Make it work well on platforms where streamed audio is disabled (this no longer uses streamed audio)
- [x] Fix menu lag on Old 3DS (SDL3 came in clutch)
- [x] Make sure there's no memory leaks 
- [ ] Crossfade between the tracks??? (probably not doing this one)

### DIFF
--- a/source/sdl/audio/audio.cpp
+++ b/source/sdl/audio/audio.cpp
@@ -9,6 +9,11 @@
 #ifdef __3DS__
 #include <3ds.h>
 #endif
+#ifdef __PC__
+#include <cmrc/cmrc.hpp>
+
+CMRC_DECLARE(romfs);
+#endif
 
 std::unordered_map<std::string, std::unique_ptr<SDL_Audio>> SDL_Sounds;
 std::string currentStreamedSound = "";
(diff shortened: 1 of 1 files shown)


Dogo6647:
epic 🦍

Dogo6647:
@NateXS 
> Make it work well on platforms where streamed audio is disabled [PS Vita, OGC] (it works on these platforms, it just doesn't transition between tracks smoothly)

Wait when I attempted to add music it didnt take like anything at all to load both tracks even without them being streamed, so idk whats going on... btw at the time I didnt use the SoundPlayer, I just raw called SDL to load the sounds and then play them both at the same time as normal audio and just adjust the volume depending on which menu stage the user is in, I have since lost my code implementation because I deleted it

Dogo6647:
so I'm pointing towards the fact that theres maybe an actual issue with the `SoundPlayer` implementation that makes sounds take longer to load than they should?

NateXS:
> so I'm pointing towards the fact that theres maybe an actual issue with the `SoundPlayer` implementation that makes sounds take longer to load than they should?

hm, maybe? I'll have to experiment with just using the raw SDL2 functions to see if it's any better.

> btw at the time I didnt use the SoundPlayer, I just raw called SDL to load the sounds and then play them both at the same time as normal audio and just adjust the volume depending on which menu stage the user is in

This sounds like the better way to implement this tbh, cause then I could make the transition sound smoother by adjusting the volumes smoothly

gradylink:
@NateXS you should probably come back to this now that you've rewritten the 3DS audio engine.

gradylink:
Oh @NateXS btw, I think you'll need to add support for loading audio from CMakeRC for this to work correctly with the new PC build (I mean it will work but only if launched from the SE! folder.) You can look at how I did it in the image and text classes.

gradylink (approved):
lgtm

PR #352, opened 2025-09-30
[WIP/Experiment]: Migrate to RapidJSON

NateXS:
There's a few advantages to migrating JSON libraries, namely that it's written in C so there's more compatibility with older consoles (PSP, OG Xbox, etc). Also that it uses less RAM, and loads things a bit faster. 

TO-DO list:

- [x] Remove the current JSON library
- [x] Get it to compile with RapidJSON
- [x] Get it loading and playing projects
- [ ] Get it working with cloud variables (untested, but this _might_ need mistpp to migrate as well)
- [ ] Fix merge conflicts 💀

### DIFF
--- a/make/Makefile_switch
+++ b/make/Makefile_switch
@@ -13,8 +13,8 @@ PKGCONF := $(DEVKITPRO)/portlibs/switch/bin/aarch64-none-elf-pkg-config
 
 TARGET		:=	scratch-nx
 BUILD		:=	build/switch
-SOURCES		:=	source source/scratch source/scratch/blocks source/scratch/menus source/sdl source/sdl/audio include/miniz include/nlohmann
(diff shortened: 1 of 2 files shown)


NateXS:
Putting this here because I'm thinking about retrying the PSP port 💀

gradylink:
> but this might need mistpp to migrate as well

Technically not needed, it would just mean that we would have both RapidJSON **and** nlohnmann/json when building with cloud variables. Also, I think the only console where we've had issues with nlohnmann/json is PSP, because the Xbox issues were Miniz (we would need to migrate to Minizip.) It shouldn't really be that hard to migrate it, though.

gradylink:
I'll work on migrating Mist++ tomorrow, I think I'll just make it a config option when building, so you can choose which one you want it to use (like how SDL2_Image can use libpng or stb_image as a backend.)

PR #353, opened 2025-09-30
[Runtime]: Add ability to run headless

NateXS:
This pull request adds the ability to use the runtime headless in the command line, without using the SDL2 backend (PC only).

Should be ready to go, just wondering if my implementation is a good way of going about it (making a `headless` folder in `source/`, adding new files and empty functions).

### DIFF
--- a/source/scratch/unzip.cpp
+++ b/source/scratch/unzip.cpp
@@ -163,12 +163,13 @@ bool Unzip::load() {
     if (Unzip::projectOpened != 1)
         return false;
 #endif
-#endif
+#else
 
     // non-threaded loading
     Unzip::openScratchProject(NULL);
     if (Unzip::projectOpened != 1)
         return false;
+#endif
 
     loadInitialImages();
     return true;


source/headless/input.cpp
```diff
@@ -0,0 +1,19 @@
...
+std::vector<std::string> Input::inputButtons;
+std::map<std::string, std::string> Input::inputControls;
+int Input::keyHeldFrames = 0;
+
+std::vector<int> Input::getTouchPosition() {
+    return {0, 0};
+}
+
+void Input::getInput() {
+}
```
gradylink, line 15: We should still be able to handle keyboard inputs.

source/headless/input.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+std::string Input::getUsername() {
+    return "";
+}
```
gradylink, line 19: This should return the name of the system user (tbh the same on SDL2 PC)

source/headless/keyboard.cpp
```diff
@@ -0,0 +1,5 @@
+#include "../scratch/keyboard.hpp"
+
+std::string Keyboard::openKeyboard(const char *hintText) {
+    return "";
+}
```
gradylink, line 10: Again, we should handle keyboard input

source/scratch/unzip.cpp
```diff
@@ -41,38 +41,43 @@ int Unzip::openFile(std::ifstream *file) {
...
-
-                    Log::log("Normal .sb3 project in SD card ");
-                    file->open(OS::getScratchFolderLocation() + filePath, std::ios::binary | std::ios::ate);
-                    if (!(*file)) {
-
-                        Log::logError("Couldnt find file. jinkies.");
-                        Log::logWarning(filePath);
-                        return 0;
-                    }
+            file->open(OS::getScratchFolderLocation() + "autoload_project.sb3", std::ios::binary | std::ios::ate);
```
gradylink: I think a better solution would be an `autoload.txt` with the name of the project, just to yk not conflict with project names at all.

source/scratch/value.hpp
```diff
@@ -52,6 +52,7 @@ class Value {
             break;
         case ValueType::BOOLEAN:
             stringValue = new std::string(*other.stringValue);
+            break;
```
gradylink: Technically not necessary since it's the last item, no reason for it to be part of this PR

PR #354, opened 2025-09-30
Cleanup Value class

TheKodeToad:
This isn't necessarily a good idea because it might introduce some issues but I could not help myself as it wasn't using std::variant and duplicated number parsing logic!

Having the shorter functions not be inline might not be a good idea idk


gradylink:
Thank you so much! I've been planning on doing this for a while but never got around to it.

NateXS:
Tested a few projects with this PR, I don't see any immediate issues in terms of Scratch parity. Should be good to go!

NateXS (approved):

PR #358, opened 2025-10-01
Create Makefile for the ds

Cammy43:
Create a makefile for the ds port. I'm using the PC makefile as a base and I will edit it after I start to figure out the ds port

### DIFF
--- a/make/Makefile_ds
+++ b/make/Makefile_ds
@@ -1,0 +1,115 @@
+.PHONY: all clean debug release
+
+#Im starting off with the PC makefile and I will edit it as needed.
+
+TARGET     := Scratch-ds
+BUILD      := build/ds
+SOURCES    := source source/scratch source/scratch/blocks source/scratch/menus source/sdl source/sdl/audio include/miniz include/nlohmann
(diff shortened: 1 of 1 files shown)


gradylink:
feel free to try! Also just a makefile is not enough, please follow the guidelines listed here:

- https://github.com/ScratchEverywhere/ScratchEverywhere/blob/main/CONTRIBUTING.md (dev stuff is near the bottom)
- https://github.com/ScratchEverywhere/ScratchEverywhere/blob/main/PORTING.md

Also, I would not recommend basing your makefile after the pc one, for all of the other ports they've been based on examples for the specific platform, so for DS that would be from here: https://github.com/devkitPro/nds-examples

Cammy43:
> Also, I would not recommend basing your makefile after the pc one, for all of the other ports they've been based on examples for the specific platform, so for DS that would be from here: https://github.com/devkitPro/nds-examples

Ah, I see. This is very useful and I'll be sure to check out all of the examples. I'm new to makefiles, so thank you for helping me get started.

NateXS:
I actually started a DS port yesterday to see if it was viable, and actually got it compiling funny enough. I can open a PR with what I did so far and we can collaborate there if you want?

PR #361, opened 2025-10-01
Sync main

Cammy43:


source/scratch/os.cpp
```diff
@@ -113,6 +113,8 @@ std::string OS::getScratchFolderLocation() {
     return "ux0:data/scratch-vita/";
 #elif defined(__3DS__)
     return "sdmc:/3ds/scratch-everywhere/";
+#elif defined(__NDS__)
+    return "sdmc:/3ds/scratch-everywhere/";
```
gradylink, line 118: This should not be the same as 3ds, you should see how sd card reading is handled for DS and what other homebrew apps do.

source/scratch/os.cpp
```diff
@@ -14,6 +14,10 @@
 #include <whb/sdcard.h>
 #endif
 
+#ifdef __NDS__
+#include <nds.h>
+#endif
+
```
gradylink, line 21: This does nothing on its own.

source/scratch/os.cpp
Cammy43: Yeah, looks like I forgot to change that lol. I was going to set that to "/scratch-everywhere/"

Cammy43:
Im closing this pull request now since NateXS has given us code to work with.

PR #362, opened 2025-10-01
Initial DS Port!

NateXS:
I originally wasn't going to make this public because I don't know how far I'll go with this, but after seeing #358 pop up, I decided to just make it public since this is slightly further along.

After seeing how running Scratch Everywhere! headless only takes up ~600KB of RAM, I decided to try and mess around with getting it to compile on DS for funsies.

So I stole a Makefile from one of PolyMars's DS games (which I think is just a DevkitPro example Makefile anyway), did a few edits and surprisingly it actually compiled...

To-do list:

- [x] Makefile
   - [x] Get it to compile
   - [ ] Fix compile warnings
- [x] Get RomFS working
- [x] Loading Projects
    - [x] RomFS unzipped project
    - [x] RomFS .sb3 project
    - [x] Non-embedded project
- [x] Main Menu
 - [x] Renderer
   - [x] Headless
   - [x] Show prints
   - [x] Sprite rendering
      - [x] Render sprites as squares
      - [x] Sprite image renderer
         - [x] Convert images to RGBA
         - [x] Convert RGBA to PAL8
         - [x] Works with unpacked projects
         - [x] Works with .sb3 projects
         - [x] Support vector images
         - [ ] ~~Make new `.swzl` image format~~ (No longer seems to really be necessary)
   - [x] Audio
   - [x] Input
      - [x] Buttons
      - [x] Touch screen / mouse
   - [ ] Bottom screen support (Bottom screen only is possible, but both screens simultaneously is not feasible as I would have to write 2 seperate renderers)
- [ ] Cloud Variables (yeeaaah we are NOT doing this one 😭)
- [x] Performance test (For the hardware, its pretty good, but image and audio loading is quite slow)
- [x] Dockerfile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ romfs/sample.ogg
 romfs/project.sb3_FILES/project.json
 Scratch-pc
 scratch-wiiu/
+scratch-wasm/
 scratch-wii.zip
 
 # Stored Files
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -43,6 +43,7 @@ body:
         - Switch
         - Vita
         - PC
+        - Web
     validations:
       required: true
   - type: textarea
--- a/.github/ISSUE_TEMPLATE/feature-request.yml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yml
@@ -27,6 +27,7 @@ body:
         - Switch
         - Vita
         - PC
+        - Web
     validations:
       required: false
   - type: textarea
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -18,3 +18,6 @@ Switch:
 
 Vita:
   - '### Device\(s\)\s*.*Vita'
+
+Web:
+  - '### Device\(s\)\s*.*Web'
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,8 @@ else ifeq ($(PLATFORM),switch)
 include make/Makefile_switch
 else ifeq ($(PLATFORM),vita)
 include make/Makefile_vita
+else ifeq ($(PLATFORM),wasm)
+include make/Makefile_wasm
 else ifeq ($(PLATFORM),nds)
 include make/Makefile_nds
 else
--- a/docker/Dockerfile.wasm
+++ b/docker/Dockerfile.wasm
@@ -1,0 +1,11 @@
+FROM emscripten/emsdk:latest AS main_builder
+
+WORKDIR /app
+
+COPY . .
+
+RUN make PLATFORM=wasm release
+
+FROM scratch AS exporter
+
+COPY --from=main_builder /app/build/wasm/release/scratch-wasm.* /scratch-wasm/
--- a/gfx/wasm/shell.html
+++ b/gfx/wasm/shell.html
@@ -1,0 +1,71 @@
+<!DOCTYPE html>
+<html lang="en-us">
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <title>Loading...</title>
+    <style>
+      * {
+        margin: 0;
+      }
+
+      html,
+      body {
+        width: 100%;
+        height: 100%;
+      }
+
+      canvas {
+        width: 100%;
+        height: 100%;
+        display: block;
+      }
+    </style>
+  </head>
+  <body>
+    <canvas
+      id="canvas"
+      oncontextmenu="event.preventDefault()"
+      tabindex="-1"
+    ></canvas>
+    <script type="text/javascript">
+      const resizeCanvas = () => {
+        const canvas = document.getElementById("canvas");
+        canvas.width = window.innerWidth;
+        canvas.height = window.innerHeight;
+      };
+
+      resizeCanvas();
+
+      Module = {
+        preRun: [],
+        postRun: [],
+        canvas: (() => {
+          const canvas = document.getElementById("canvas");
+          canvas.addEventListener(
+            "webglcontextlost",
+            (e) => {
+              alert(
+                "WebGL context lost. You will need to reload the page.",
+              );
+              e.preventDefault();
+            },
+            false,
+          );
+          return canvas;
+        })(),
+        print: (() => {
+          return (text) => {
+            if (arguments.length > 1) {
+              text = Array.prototype.slice.call(arguments).join(" ");
+            }
+            console.log(text);
+          };
+        })(),
+      };
+
+      window.addEventListener("resize", resizeCanvas);
+    </script>
+    {{{ SCRIPT }}}
+  </body>
+</html>
--- a/include/emscripten_browser_file.h
+++ b/include/emscripten_browser_file.h
@@ -1,0 +1,124 @@
+#pragma once
+
+#include <string>
+#include <string_view>
+#include <emscripten.h>
+
+#define _EM_JS_INLINE(ret, c_name, js_name, params, code)                          \
+  extern "C" {                                                                     \
+  ret c_name params EM_IMPORT(js_name);                                            \
+  EMSCRIPTEN_KEEPALIVE                                                             \
+  __attribute__((section("em_js"), aligned(1))) inline char __em_js__##js_name[] = \
+    #params "<::>" code;                                                           \
+  }
+
+#define EM_JS_INLINE(ret, name, params, ...) _EM_JS_INLINE(ret, name, name, params, #__VA_ARGS__)
+
+namespace emscripten_browser_file {
+
+/////////////////////////////////// Interface //////////////////////////////////
+
+using upload_handler = void(*)(std::string const&, std::string const&, std::string_view buffer, void*);
+
+inline void upload(std::string const &accept_types, upload_handler callback, void *callback_data = nullptr);
+inline void download(std::string const &filename, std::string const &mime_type, std::string_view buffer);
+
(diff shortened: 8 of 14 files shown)


NateXS:
🥹
image

Cammy43:
Wow! This is a LOT more to work with. It even already boots! I'm going to see about all of those renderer issues.

Cammy43:
I was looking through the devkitpro's nds-examples and I came across a library called "gl2d". Im interested because we don't have sprite rendering on the DS and all of source/nds/image.cpp's functions are empty. Gl2d has image rendering capabilities among other things. I think we could use gl2d instead of sdl. 

Link to examples: https://github.com/devkitPro/nds-examples/tree/master/Graphics/gl2d

gradylink:
> - [ ]  Fix compile warnings

Let me guess, it's compile warnings from Miniz or nlohnmann/json. They happen on some other niche platforms, we can just ignore them as long as they aren't coming from our code.

> - [ ] Cloud Variables (yeeaaah we are NOT doing this one 😭)

hmmm...

EDIT: OK actually no.

> - [ ] Sprite image renderer (this is definitely gonna be tough)

More sizzle magic??!??!

Cammy43:
I have found a way better graphics library that also supports audio. Its called nightFoxLib. NFL is way simpler and easier to read than gl2d. All of is features seem to be inline with what we need. Check it out here: https://github.com/knightfox75/nds_nflib

NateXS:
> I have found a way better graphics library that also supports audio. Its called nightFoxLib. NFL is way simpler and easier to read than gl2d. All of is features seem to be inline with what we need. Check it out here: https://github.com/knightfox75/nds_nflib

yup, I was planning on using nightfox lib, but I'm wondering if it's worth it to use gl2d anyways, in case NFL has a ton of overhead. I guess we can try NFL first and if we hit any limits, try out gl2d?

Cammy43:
> > I have found a way better graphics library that also supports audio. Its called nightFoxLib. NFL is way simpler and easier to read than gl2d. All of is features seem to be inline with what we need. Check it out here: https://github.com/knightfox75/nds_nflib
> 
> yup, I was planning on using nightfox lib, but I'm wondering if it's worth it to use gl2d anyways, in case NFL has a ton of overhead. I guess we can try NFL first and if we hit any limits, try out gl2d?

Sounds great, i'll get started.

Cammy43:
I have notice nfl kinda wants me to use something called "nitroFS"...
I'm going to look into that also because we might be forced to use it for things like loading images.

edit: I found a workaround.

RunTheCoins:
> I have notice nfl kinda wants me to use something called "nitroFS"... I'm going to look into that also because we might be forced to use it for things like loading images.
> 
> edit: I found a workaround.

NitroFS probably wouldn't hurt to use honestly, I'd still suggest looking into it

NateXS:
Just found out we also need a DS implementation of the `Timer` class, since currently projects are running at exactly 1 FPS (removing the checkframerate() function makes it run normally)

NateXS:
Got button input working!
image
Surprisingly the 3DS and DS input API is _really_ similar, so most of this was a copy-paste job

Cammy43:
> NitroFS probably wouldn't hurt to use honestly, I'd still suggest looking into it

The only thing is that the assets need to be embedded with the project.
But if this isn't a problem sure

Cammy43:
> Just found out we also need a DS implementation of the `Timer` class, since currently projects are running at exactly 1 FPS (removing the checkframerate() function makes it run normally)

I can probably figure this out, I need a break from the renderer. It's hard to wrap my head around.

Edit: I figured it out.
Since SDL is supported on the DS, we can use the getTicks() function and similar for timekeeping.

Cammy43:
Working on the renderer is hard without really knowing how it works. I'm working on the nds/renderer.cpp and nds/image.cpp files.   I have looked through almost every file in the repo and I have gotten a basic understanding. Could someone explain how it works and how the rest of the code interacts with it? It would really help speed up the process and I could be more helpful.

Cammy43:
> * [ ]  Cloud Variables (yeeaaah we are NOT doing this one 😭)

I've come up with a really easy implementation for cloud variables. Inside the scratch-everywhere directory there can be a directory (ex. "cloudVars") that has a directory for every project that uses cloud variables. Then, for each project, just store every cloud variable inside the project's cloudVars directory.

NateXS:
We've got sprites rendering!

https://github.com/user-attachments/assets/7e46e5f6-b539-45f2-a8f1-16797a008179

winbamstudios:
Can you provide the NDS file? I should be able to test on my DSi.

gradylink:
> Can you provide the NDS file? I should be able to test on my DSi.

You'll need to compile it yourself, this is WIP for a reason.

NateXS:
The main menu is working!

image

If anyone looking at this has a DS / DS Lite / DSi, I need to figure out if the project menu works correctly on there. 

Download this nds file: [scratch-ds.zip](https://github.com/user-attachments/files/22783028/scratch-ds.zip) and run it on your DS, putting a Scratch project in `sd card:/scratch-ds/`. If you're able to load it, then cool! It probably won't run, but I just need so see if SE can find the projects at all.

StarryPen:
The menu loads but in the username settings, the text doesn't
It also says the scratch-ds folder doesn't exist

the-smartest-cat:
tried to run it in twilight menu++ since my flashcarts don't seem to be working in general. It returned with an error message (Error: NitroFS Could not initialize!)

KyleKart:
> tried to run it in twilight menu++ since my flashcarts don't seem to be working in general. It returned with an error message (Error: NitroFS Could not initialize!)

I tried running it in whatever is on my flashcard on my DSLite and got the same.

rttyg46305Unj:
Very impressive!

.github/workflows/nightly-nds.yml
```diff
@@ -0,0 +1,39 @@
+name: DS Nightly Build
+
+on:
+  push:
+    paths:
+      - make/Makefile_nds
+      - source/scratch/**
+      - source/sdl/**
```
gradylink: This should be `source/nds/**`

source/nds/audio.cpp
```diff
@@ -0,0 +1,379 @@
...
+    size_t multiplier = 0;
+
+    if (format == MM_STREAM_8BIT_MONO)
+        multiplier = 1;
+    else if (format == MM_STREAM_8BIT_STEREO)
+        multiplier = 2;
+    else if (format == MM_STREAM_16BIT_MONO)
+        multiplier = 2;
+    else if (format == MM_STREAM_16BIT_STEREO)
+        multiplier = 4;
```
gradylink: I think this should be a `switch` statement.

source/nds/audio.cpp
```diff
@@ -0,0 +1,379 @@
...
+            break;
+        }
+    }
+}
+
+void NDS_Audio::streamingFillBuffer(bool force_fill, bool restartSound) {
+    if (!force_fill) {
+        if (stream_buffer_in == stream_buffer_out)
+            return;
+    }
```
gradylink: Shouldn't this just use an `&&`.

source/nds/audio.cpp
```diff
@@ -0,0 +1,379 @@
...
+        if (bitsPerSample == 8)
+            return MM_STREAM_8BIT_MONO;
+        else
+            return MM_STREAM_16BIT_MONO;
+    } else if (numChannels == 2) {
+        if (bitsPerSample == 8)
+            return MM_STREAM_8BIT_STEREO;
+        else
+            return MM_STREAM_16BIT_STEREO;
+    }
```
gradylink, line 147: All the `else`s here are unnecessary and arguably make the code harder to read (at least in my opinion.)

source/nds/audio.cpp
```diff
@@ -0,0 +1,379 @@
...
+            };
+        mmStreamOpen(&stream);
+        audio.isPlaying = true;
+        std::string baseName = fileName.substr(fileName.find_last_of("/\\") + 1);
+        NDS_Sounds[baseName] = std::move(audio);
+        NDS_Sounds[baseName].id = baseName;
+        return true;
+    }
+
+    return false;
```
gradylink: Wouldn't a `if (!streamed) return false;` then the code in the `if` work better here.

source/nds/image.cpp
```diff
@@ -0,0 +1,556 @@
...
+        //     renderScale = (renderScale * data.scaleX) >> 12;
+        // }
+
+        glSpriteScale(RenderX, RenderY, renderScale, GL_FLIP_NONE, image);
+        data.freeTimer = data.maxFreeTimer;
+    }
+}
+
+void Image::renderNineslice(double xPos, double yPos, double width, double height, double padding, bool centered) {
+    // we are NOT doing nine-slice rendering on the DS....
```
gradylink, line 83: I might be willing to do it, btw, I'll have to see how DS handles sub-textures.

EDIT:
```c
typedef struct
{
    int width;     ///< Width of the Sprite
    int height;    ///< Height of the Sprite
    int u_off;     ///< S texture offset
    int v_off;     ///< T texture offset

    /// Texture handle (used in glDeleteTextures())
    ///
    /// The texture handle in VRAM (returned by glGenTextures()). ie. This
    /// references the actual texture stored in VRAM.
    int textureID;

} glImage;
```

Yeah, this is definitely possible, it'll use a similar approach to 3DS.

.gitignore
```diff
@@ -58,6 +58,7 @@ main.code-workspace
 *.vpk
 romfs/sample.ogg
 romfs/project.sb3_FILES/project.json
+scratch-ds.nds
```
gradylink: To match the other platforms, this should be `*.nds`.

PR #364, opened 2025-10-01
Added more nightfoxlib stuff to Feat/ds port

Cammy43:
There are some unfinished functions in nds/image.cpp but those are there so I remember to figure out how to use them.


gradylink:
Also, it seems you didn't run `clang-format` before saving this, please do that before committing.

PR #378, opened 2025-10-02
Feat/ds port Add swzl image library!

Cammy43:
SWZL is now semi-functional.

### DIFF
--- a/docker/Dockerfile.windows
+++ b/docker/Dockerfile.windows
@@ -1,0 +1,62 @@
+FROM debian:bookworm AS main_builder
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y \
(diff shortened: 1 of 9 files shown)


source/nds/swzl.hpp
gradylink, line 1: A few things, this should be split into `swzl.cpp` and `swzl.hpp`, it also needs to be formatted.

gradylink:
I'm not going to merge this until it is done and confirmed to be faster/less ram.

Cammy43:
OK, I'll figure out the glTexture stuff tomorrow. I'm not too sure about it being faster, but lighter on ram and storage. I'll try to get it to about the same speed.

Dogo6647:
no swzl?

PR #380, opened 2025-10-03
Nintendo Gamecube Logo on Control Graphic

ikyih:
Why wasn't this here before???
If this gets accepted i'll do wii too


OceanIsEndless:
I'm not a contributor, but be sure to remove the hidden .DS_Store files from your pull! Macs put those there automatically, but they shouldn't be committed to the repo. Alternatively you could add ".DS_Store" to the .gitignore to make sure they are excluded in future pulls.

ikyih:
i know but it doesnt let me

gradylink:
Not really needed and um, might as well avoid copyright.

PR #385, opened 2025-10-04
3D Banner! (3DS)

Giolaboman:
3DBanner_Preview

**Made a 3D banner for the 3DS!** 

Makefile is updated to use my already-compiled banner.cgfx that contains the 3D banner. Ran it successfully! The gif above shows how it looks :)

Also included: I changed the banner sound (taken from the Scratch Library), but you don't have to use it. Just thought it would work. 

Enjoy!! :)


3DBanner_Preview.gif
gradylink, line 1: I don't think this file needs to be in here.

make/Makefile_3ds
```diff
@@ -341,7 +344,7 @@ $(OUTPUT).elf	:	$(OFILES)
 # CIA build (makerom + bannertool)
 #---------------------------------------------------------------------------------
 banner.bnr: $(BANNER_PNG) $(BANNER_WAV)
-	@$(BANNERTOOL) makebanner -i $(BANNER_PNG) -a $(BANNER_WAV) -o $@
+	@$(BANNERTOOL) makebanner -ci $(BANNER_3D) -a $(BANNER_WAV) -o $@
```
gradylink, line 294: Maybe we could make 2d/3d banner a compilation option.

3DBanner_Preview.gif
Giolaboman: nah it doesn't, just added it there for you guys to see 😭

gradylink: You should remove it then.

Giolaboman: Removed!

make/Makefile_3ds
gradylink: This should probably be done, since the old assets are still around.

gradylink:
You pushed your new PR to this one, so I'll just close this.

PR #386, opened 2025-10-04
[WIP] Switch PS Vita port from SDL2 to vita2d

CatsLover2006:
More native code for another platform, to reduce overhead, although the PS Vita is admittedly a bad place to start, it's just the console I'm most familiar with coding for (outside the 3DS, which... well...)<br>
This PR, as of submitting it, is more of a proof-of-concept than anything else. I haven't actually tested my build, at all.<br>
Some notable things I've done in the current version:

- Use vita2d over SDL2 (the entire point of the PR)
- Use some cursed programming techniques to speed up a data copy, which probably shouldn't happen in the first place since I prefer a different and much simpler image loading routine
- Bang my head against the Makefile for a solid half hour
- Poorly implement the ghost effect
- Not implement the brightness effect
- Copy the SDL2 audio setup (I'll write a different mixer at some point)
- Not implement vector images (although set up a framework to allow them to be drawn as vectors, regardless of platform)
- Not implement loading images from SB3 files
- Not implement... freeing leftover image data (granted, they also stay referenced, and will be still used if the need pops up, but still... definitely need to finish that)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ scratch-everywhere/
 
 main.code-workspace
 
+# macOS
+.DS_Store
+
 # Prerequisites
 *.d
 
@@ -64,4 +67,3 @@ scratch-wii.zip
 
 # Stored Files
 cloud-username.txt
-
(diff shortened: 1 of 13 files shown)


gradylink:
Closing this as stale, feel free to open a new PR if you decide to continue working on this.

PR #393, opened 2025-10-07
Say & Think blocks

roccopm:
## Description of changes
Complete implementation of Say and Think blocks, including timed speech (e.g., "Say `x` for `y` seconds").

### Known limitations (for now)
- Does not yet have full feature parity with Scratch 3. One example is, speech bubbles do not "remember" which direction they face in this implementation, they simply check what side of the screen they're on (relative to the middle) and set position based on that.
- Position relative to sprite is calculated in the same way that collision is, which isn't perfect and may cause the speech to overlap with the sprites.
- Visuals are _not_ the same between platforms due to hardware limitations.

### PC (SDL2)
image

### 3DS
image

### DS
image


source/sdl/speech_manager_sdl.cpp
```diff
@@ -0,0 +1,176 @@
...
+void SpeechManagerSDL::renderSlice(Image &image, int destX, int destY, int destW, int destH, int srcX, int srcY, int srcW, int srcH) {
+    if (images.find(image.imageId) != images.end()) {
+        SDL_Image *sdlImage = images[image.imageId];
+
+        SDL_Rect srcRect = {srcX, srcY, srcW, srcH};
+        SDL_Rect destRect = {destX, destY, destW, destH};
+
+        SDL_RenderCopy(renderer, sdlImage->spriteTexture, &srcRect, &destRect);
+    }
+}
```
gradylink: We already have nine-slice handling in the `Image` class. I don't think it existed when you started working on this though, so don't really worry about it, the same nine-slice function also exists for 3DS as well.

source/scratch/blocks/looks.cpp
```diff
@@ -3,6 +3,13 @@
 #include "image.hpp"
 #include "interpret.hpp"
 #include "math.hpp"
+
+#ifdef __3DS__
+#include "speech_manager_3ds.hpp"
+#else
+#include "speech_manager_sdl.hpp"
+#endif
+
```
NateXS, line 16: This, and other #ifdefs of this nature, should be using #elif defined(SDL_BUILD) instead of #else, to anticipate supporting other renderers.

roccopm: Absolutely, that's an oversight on my part. I'll sort it, thanks

roccopm: > This, and other #ifdefs of this nature, should be using #elif defined(SDL_BUILD) instead of #else, to anticipate supporting other renderers.

I've resolved this by adding that elif defined(SDL_BUILD) to all instances, and using the basic (not platform specific) SpeechManager class. If a new renderer is added now, there won't be any issues if speech bubbles are omitted from its includes / loop. I've tested this by commenting out all the SDL checks (so it hits the #else) and building for SDL.

See latest commit

source/sdl/speech_manager_sdl.cpp
roccopm: Thanks for letting me know, I've just swapped it over to use your implementation

roccopm:
I've noticed that on 3DS, rendering these speech bubbles with nine-slice rendering immediately tanks the framerate if there's more than one at a time. I suspect this is due to limited VRAM. This will almost certainly be an issue on smaller SDL2 consoles also, so I'm going through them to test their performance.

As a fix, I think it's probably best to skip the speech bubble graphic entirely and just render the black text on a white rectangle for these more limited devices. If it's a "Think" and not a "Say", it should probably just render the text in italic. I can put #ifdefs in `speech_manager_sdl.cpp` to decide which rendering method to use based on console and how powerful it is. Limiting the number of speech bubbles that can be displayed at once is also an option.

Does anyone have any thoughts on this?

gradylink:
> I've noticed that on 3DS, rendering these speech bubbles with nine-slice rendering immediately tanks the framerate if there's more than one at a time. I suspect this is due to limited VRAM. This will almost certainly be an issue on smaller SDL2 consoles also, so I'm going through them to test their performance.

My 3DS nine-slice implementation actually doesn't increase (V)RAM usage at all. I think it's just the fact that it needs to do a ton of complicated calculations for each slice.

Instead of doing nine-slice we could use rounded rectangle rendering, IDK how to do it with Citro2D on 3DS (we could ask in the hb discord), but for SDL2 we can take Clay's implementations: https://github.com/nicbarker/clay/blob/588b93196cc7a182a024a9ef08ba8e352904a1bd/renderers/SDL2/clay_renderer_SDL2.c#L151 (for the outline), and https://github.com/nicbarker/clay/blob/588b93196cc7a182a024a9ef08ba8e352904a1bd/renderers/SDL2/clay_renderer_SDL2.c#L47 (for the inside white background.)

roccopm:
**Quick update due to inactivity:** This isn't abandoned, I've just been a little preoccupied. Planning to get this PR finalised this weekend and have speech performant and looking good on slower hardware.

source/renderers/citro2d/speech_manager_c2d.cpp
```diff
@@ -0,0 +1,165 @@
...
+        origBottom};
+
+    C2D_ImageTint tinty;
+    C2D_AlphaImageTint(&tinty, speechIndicatorImage->opacity);
+
+    // Render using the clipped subtexture
+    float scaleX = static_cast<float>(indicatorSize) / static_cast<float>(halfWidth);
+    float scaleY = static_cast<float>(indicatorSize) / static_cast<float>(fullHeight);
+    float adjustedY = static_cast<float>(indicatorY) + (static_cast<float>(fullHeight) / 2.0f) - (static_cast<float>(indicatorSize) / 2.0f);
+    C2D_DrawImageAt({image.tex, &clippedSubtex}, static_cast<float>(indicatorX), adjustedY, 1, &tinty, scaleX, scaleY);
```
roccopm, line 162: I tried using `renderSubrect` from the image class for this but couldn't get it to work right so I implemented it manually. I believe this is because renderSubrect doesn't work with padded images (they get filled in during `loadImageFromFile` to ensure power-of-2), and my image is 32x16, so it was calculating the UVs wrong for where my actual textures were within that space. This implementation definitely works, though.

gradylink: It should account for padding, but it only accounts for padding on the right and bottom, not the top or left.

source/sdl/speech_text_sdl.cpp
```diff
@@ -8,7 +8,9 @@ SpeechTextObjectSDL::SpeechTextObjectSDL(const std::string &text, int maxWidth)
     : TextObjectSDL(text, 0, 0), SpeechText(text, maxWidth) {
     setColor(0x00);
     setCenterAligned(false); // easier for positioning logic
-    font = TTF_OpenFont("gfx/menu/Arialn.ttf", 16);
+    
+    std::string fontPath = OS::getRomFSLocation() + "gfx/menu/Arialn.ttf";
+    font = TTF_OpenFont(fontPath.c_str(), 16);
```
gradylink: You don't need an intermediate here, you can just do:
```suggestion
    font = TTF_OpenFont((OS::getRomFSLocation() + "gfx/menu/Arialn.ttf").c_str(), 16);
```

roccopm: I've made the change. Thanks!

roccopm:
This needs tested on real hardware (especially 3DS and DS) by someone else as I don't have any, but other than that, I think it's ready to review!

If there are any glaring compatibility, performance or code structure issues then let me know and I'll draft it again if needed. If anyone has thoughts on visual choices then feel free to question me and discuss those as well, I'm not precious about any of it and I'm happy to change things around.

source/renderers/gl2d/text_gl2d.hpp
```diff
@@ -12,23 +12,26 @@ typedef struct {
     int textureID;
     int atlasWidth;
     int atlasHeight;
-    int fontPixels;
```
roccopm, line 15: Due to making the NDS text font size changeable, I renamed "fontPixels" to "fontSize" for clarity when text is created. If I shouldn't have made this naming convention change, let me know and I'll switch it back

roccopm:
## Performance test results
I have tested PC (SDL2), 3DS and DS for performance with a high number of speech bubbles. I only took them up to the 600 / 700 range as I don't think any project is ever going to have any more than that unless it's actively trying to hurt performance (like this torture test is). FPS stayed high and consistent the entire time, no drops on any of the platforms even at these high counts. The only issue is that on DS, after a number of speech bubbles are rendered, it starts refusing to render any new text (RAM and rendering constraints, I imagine). I don't think this is a major issue as it only started happening after like 50 bubbles were being rendered at once, which is still an unrealistic amount to reach by accident.

### PC (SDL2)
image

### 3DS
image

### DS
image

source/renderers/gl2d/text_gl2d.hpp
gradylink: Sounds good!

source/3ds/speech_manager_3ds.cpp
```diff
@@ -0,0 +1,165 @@
+#include "speech_manager_3ds.hpp"
+#include "image.hpp"
+#include "interpret.hpp"
+#include <3ds.h>
+
+void renderSubrect(C2D_Image img, uint16_t srcX, uint16_t srcY, uint16_t srcW, uint16_t srcH, float destX, float destY, float destW, float destH, C2D_ImageTint *tint);
```
gradylink: If you're not going to be using `renderSubrect`, then you shouldn't define it here. Also, if you were using it, it should be `extern`.

roccopm: This was left over from when I was trying to use it, just forgot to remove it. My bad! Fixed in latest commit.

source/3ds/speech_manager_3ds.cpp
```diff
@@ -0,0 +1,163 @@
...
+    speechObjects[sprite] = std::make_unique<SpeechTextObject3DS>(message, 100);
+}
+
+void SpeechManager3DS::render() {
+    // Ensure images are loaded (they may have been cleaned up)
+    ensureImagesLoaded();
+
+    // Get screen dimensions and scale so speech size aligns with resolution
+    const int SCREEN_WIDTH = 400;
+    const int SCREEN_HEIGHT = 240;
```
NateXS: These should be using Render::getWidth() and Render::getHeight() so it sets the correct resolution based on if the bottom screen is in use or not.

source/3ds/speech_manager_3ds.cpp
```diff
@@ -0,0 +1,163 @@
...
+    if (styleIt == speechStyles.end()) return;
+
+    std::string style = styleIt->second;
+
+    if (!speechIndicatorImage || speechIndicatorImage->imageId.empty()) return;
+    if (images.find(speechIndicatorImage->imageId) == images.end()) return;
+
+    int cornerSize = static_cast<int>(8 * scale);
+    int indicatorSize = static_cast<int>(16 * scale);
+    const int SCREEN_WIDTH = 400;
```
NateXS: should be Render::getWidth()

NateXS (changes requested):

source/3ds/speech_manager_3ds.cpp
roccopm: Fixed both instances of this 👍🏼

roccopm:
There is currently an issue with the SDL2 builds where scaling up the window causes the 9-slice rendering (used on the speech bubble graphic) to break. This isn't an issue with the speech bubble rendering _specifically_, but with the nine-slice rendering method itself. I believe it's best to put this fix in a follow-on PR as it's not directly related to speech and could have knock-on effects with the main menu layout which would need resolved too.

Here is what that looks like at its worst:
image

Happy to discuss further.

gradylink:
Making this a draft since it needs SDL1 and SDL3 support which should be pretty easy, you should be able to pretty much just copy and paste the SDL2 code and change a few things, you can refer to the SDL3 migration guide.

roccopm:
### SDL1
Had to make some compromises on graphics due to the nine-slice not being there. Not going to attempt to implement that in this PR as, as far as I can tell from a quick bit of googling, SDL1 does not support GPU-accelerated scaling with SDL_BlitSurface. It looks like this is why nine-slice rendering wasn't implemented for this platform in the first place. For this reason, I have frankensteined together the rendering from both SDL2 and 3DS implementations to make this:
Screenshot From 2025-12-24 00-25-28
image
The current fallback for nine-slice is just to stretch the image which looks worse than just using simple rects in this case.

### SDL3
The SDL3 version, of course, needed no compromises. It was just syntax changes.
image

gradylink (approved):

CMakeLists.txt
```diff
@@ -537,11 +537,11 @@ if(NOT NINTENDO_SWITCH AND NOT PSP AND NOT VITA AND NOT WEBOS)
...
 
 if(SE_RENDERER STREQUAL "sdl1")
-	target_compile_definitions(scratch-everywhere PRIVATE RENDERER_SDL1)
+	target_compile_definitions(scratch-everywhere PRIVATE RENDERER_SDL1 SDL_BUILD)
 elseif(SE_RENDERER STREQUAL "sdl2")
-	target_compile_definitions(scratch-everywhere PRIVATE RENDERER_SDL2)
+	target_compile_definitions(scratch-everywhere PRIVATE RENDERER_SDL2 SDL_BUILD)
 elseif(SE_RENDERER STREQUAL "sdl3")
-	target_compile_definitions(scratch-everywhere PRIVATE RENDERER_SDL3)
+	target_compile_definitions(scratch-everywhere PRIVATE RENDERER_SDL3 SDL_BUILD)
```
roccopm: Added this so you can just check if it's an SDL build for imports without checking SDL 1 2 and 3 because the filenames are the same across versions anyway. This is just a convenience thing; I can see other places in the codebase are just checking all 3 so if it's better to omit this for consistency then I'd be happy to revert it. Just let me know 👍🏼

gradylink: It's probably best to omit this for consistency, especially with how #473 is going to handle new renderers.

roccopm: Reverted in latest commit

roccopm:
During amendments, I updated SDL1 to match 3DS more closely with the simplified images. The SDL icons weren't meant to go with the rect. Here's what SDL1 looks like now:
image

roccopm:
Drafted until merge conflicts resolved and feature re-tested following large refactor PR

roccopm:
Replaced by #538

Issue #397, opened 2025-10-07
[BUG] (minor) Main menu splash text gets larger over time


roccopm:
### Describe the crash or bug

If you leave the main menu running for several minutes, the splash text gradually gets larger and larger. I assume it's related to the zoom in / zoom out animation.

I noticed this when I accidentally left the project running in the main menu in the background, and when I found it, the splash text was huge.

I've tested this on a PC build and on dolphin-emu (emulation for Wii build), and it only seemed to happen on the Wii, though the PC build may have just been taking much longer due to stability from running on intended hardware.

### Scratch Project

N/A

### When opening the app
<img width="641" height="484" alt="Image" src="https://github.com/user-attachments/assets/257356b1-6219-42d0-b10e-6789d8bef667" />

### After several minutes
<img width="646" height="480" alt="Image" src="https://github.com/user-attachments/assets/3f0efe07-00c1-4863-97d3-847925795954" />

### Device(s)

PC

### Additional context

I don't want to admit how long I spent clicking in and out of the projects menu to get the "Woo, softwaregore!" splash for the second screenshot. The irony of it being on the first one was not lost on me..

### Pre-submission Checks

- [x] I have searched existing issues and confirmed this bug has not been reported before.
- [x] I have checked the known issues/roadmap and this is not a known limitation or planned fix.

roccopm:
Update - This definitely does happen on PC build also, just **much** slower. Notice how the splash text has expanded beyond the sides of the logo:
<img width="545" height="427" alt="Image" src="https://github.com/user-attachments/assets/986e6d6d-2bf0-4e09-8692-9aa09657320e" />

gradylink:
Oh, we thought this issue was specific to the WIP PS4 port, I guess not.

PR #399, opened 2025-10-08
[BUG] Fix title and splash text animation issues

roccopm:
Fixes issue #397 

## Description of issue
If you leave the main menu running for several minutes, the splash text gradually gets larger and larger. I assume it's related to the zoom in / zoom out animation.

I noticed this when I accidentally left the project running in the main menu in the background, and when I found it, the splash text was huge.

## Changes made
Fixed splash text size issue by adding splashZoom (samples the sine wave to determine text size) to the **original** splash size (stored in new var "splashTextOriginalScale" that gets set on init) to avoid tiny floating-point errors accumulating in the scale. Essentially, setting the splashText->scale to 1 + sine wave, instead of sampling the sine wave and adding that directly to splashText->scale.

While testing this, I also noticed that the bobbing effect starts twitching and skipping frames when the main menu has been open for a while. I found that this was due to precision issues when the elapsed time being used for the animation grew too large, so I fixed that in this PR as well.

## Testing
To test this, I left the dolphin-emu emulator running in the background for half an hour or so while doing other tasks, which is how I recreated the issue in #397. This time, when I came back, the text was the correct size.

### DIFF
--- a/source/scratch/menus/mainMenu.cpp
+++ b/source/scratch/menus/mainMenu.cpp
@@ -6,6 +6,7 @@
 #include "projectMenu.hpp"
 #include "settingsMenu.hpp"
 #include <cctype>
+#include <cmath>
 
 Menu::~Menu() = default;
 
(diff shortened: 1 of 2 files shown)


NateXS:
Looks good to me!

PR #409, opened 2025-10-15
[3DS] Migrate to SDL3 for audio

NateXS:
Pros:
- Sounds will work when using `.cia` 
- Streamed sounds work again
- Old 3DS performance is way better with streamed audio
- All sounds are streamed (no more needing to put audio in the stage, at least when targetting the 3DS)
- Sounds load a bit faster

Cons:
- Audio gets noticeably slowed down / starts cracking on Old 3DS when a bunch of sounds are playing at once, or when the game is lagging.
- Overall project performance is lower on Old 3DS when playing audio. (It's a lot faster than when playing SDL2 streamed audio, but slower than when playing SDL2 normal audio)


gradylink:
In my opinion it's close, but if we add an option to just disable audio (I think globally would be best) I feel like that partially removes the last con and then the pros outweigh the cons.

KirbySuper008:
I mean, you could try having a toggle for loading sounds or streaming everything if the streaming is what causes it to slowdown on O3DS.

make/Makefile_3ds
```diff
@@ -50,7 +50,7 @@ GFXBUILD	:=	$(ROMFS)/gfx
 ENABLE_CLOUDVARS	?=	0
 ENABLE_LOADSCREEN	  ?=	1
 ENABLE_AUDIO	?=	1
-RAM_AMOUNT		?= 72
+RAM_AMOUNT		?= 64
```
gradylink: Is this meant to be part of this pr

NateXS: woops, forgot to change that back

PR #417, opened 2025-10-21
menu when you click start or + in a project

Br0tcraft:
There's already been talk about integrating some kind of way to exit the Scratch project and return to the menu without requiring the Scratch project to contain a "stop all" block. Here's my approach to implementing such a thing.
To open the menu, simply press and hold the start button (or, on consoles without a start button, press the + button).


gradylink:
@Br0tcraft can I see a screenshot of what it looks like. Also, I think it makes more sense to do a dual or tri-button combo to prevent conflicts with projects.

Br0tcraft:
@gradylink here you go
image
It's very simple. Unfortunately, I'm not that good at design, but it should be good enough, right?
(3ds and sdl versions look identical)

Which double or triple key combination would you prefer?

gradylink:
> Which double or triple key combination would you prefer?

I think maybe `start` + `left bumper` would be the best combo because I don't think there would be very many games that need that combo.

EDIT: This is going to be fun to reimplement with the UI rewrite...

source/scratch/interpret.cpp
```diff
@@ -179,6 +180,38 @@ bool Scratch::startScratchProject() {
                 shouldStop = false;
                 return true;
             }
+            if (Input::keyHeldFrames > 30 && Input::isButtonPressed("start")) {
```
NateXS: the menu should only be accessible if projectType == UNEMBEDDED imo, so romFS projects can't access it

Dogo6647: i **absolutely** agree with this

Br0tcraft: Oh, lol, thanks, sorry, I thought I already included that.

Br0tcraft:
I think that should work?

The only thing is that the buttons for opening the menu change when you use the "change controls" function. Unfortunately, I don't understand the input class well enough. If one of you could fix that?

source/scratch/input.hpp
```diff
@@ -90,6 +90,16 @@ class Input {
...
+    static bool isButtonPressed(const std::string &buttonName) {
+
+        auto it = inputControls.find(buttonName);
+        if (it == inputControls.end()) return false;
+
+        const std::string &mappedKey = it->second;
+
+        return std::find(inputButtons.begin(), inputButtons.end(), mappedKey) != inputButtons.end();
+    }
+
```
gradylink, line 102: You added this but don't seem to use it.

gradylink:
> I think that should work?
> 
> The only thing is that the buttons for opening the menu change when you use the "change controls" function. Unfortunately, I don't understand the input class well enough. If one of you could fix that?

@Br0tcraft, you could either upstream the changes I made to the `Input` class in the custom extensions branch, or you could just use the function that is used for each platform in the `getInput` function.

source/scratch/input.hpp
Br0tcraft: in the interpreter.cpp script, on line 180(I thought that would give me the actual keys) I just can't seem to get it right.
I don't seem to understand the Input class well enough.

Br0tcraft:
@gradylink Could we still merge it? Then you or Natex could modify the key combination? (that would be awesome) I'm too clueless for that.

PR #423, opened 2025-10-29
Add guide to build SE! romfs projects

Br0tcraft:
I thought SE could possibly offer precise instructions to build the app with Docker (for windows)?? I put something together there. Are there any suggestions? Or is the guide even necessary?
(How and where should it be linked in the readme?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ If you would like to change the name of the app or any other information you can
 
 The recommended way to compile Scratch Everywhere! is with Docker. 
 
-We have written a step-by-step guide with pictures (for Windows) that explains the build process very simply, you can find it here: https://github.com/ScratchEverywhere/ScratchEverywhere.github.io/blob/main/docker.md
+We have written a step-by-step guide with pictures (for Windows) that explains the build process very simply, you can find it here: https://ScratchEverywhere.github.io/docker
 To compile with Docker all you need installed is Docker and Buildx.
 
 -   To compile for the **3DS**, run `docker build -f docker/Dockerfile.3ds --target exporter -o . .`.


Dogo6647:
Very necessary, thanks for taking the time to put this together!

guide/guide.md
```diff
@@ -0,0 +1,201 @@
...
+---
+
+## Step 2: Download Scratch Everywhere
+
+1. Visit the official Scratch Everywhere repository:  
+	[https://github.com/ScratchEverywhere/ScratchEverywhere](https://github.com/ScratchEverywhere/ScratchEverywhere)
+2. Click the green **“<> Code”** button.  
+   ![image](pictures/Github.png)
+3. Choose **“Download ZIP”**  
+   ![image](pictures/download.png)
```
Dogo6647: To my understanding doing this will always download the latest nightly. Maybe this part should instead instruct the user to go to the releases tab and then click the "Source code (zip)" option in the latest beta *release*, which holds a snapshot of the repo the moment the beta was released.

Dogo6647:
something else i noticed, this guide seems to be very windows-centric, which is okay as it's probably what most people who are gonna follow this guide are gonna use, and it's easy to assume linux users being linux users will probably figure stuff like this out easily but it would be nice to add instructions for other platforms. 

(maybe you could add "Unix" instructions so that you don't have to write instructions for both macOS and linux bc they both operate in very similar environments)

gradylink:
I think this makes more sense to be in the https://github.com/ScratchEverywhere/ScratchEverywhere.github.io repo, which desperately needs an update, we should really finish migrating to it soon.

juniyaaa:
kirrrrrb from the discord here. really helpful - thank you for doing this ☺️

Br0tcraft:
@Dogo6647 This shouldn't make much of a difference. Placing, unpacking, etc., works identically under Linux. The only differences are when downloading Docker (if it's not pre-installed) or opening the terminal.

Linux users should be able to open the terminal without any problems. The command to navigate to the folder is also included in the instructions. And if you select the Linux version of Docker to download, you'll be redirected to a page where everything is explained.
However, I can still take screenshots from Linux. If that's necessary?

Br0tcraft:
@gradylink I've removed the guide here and submitted a pull request to the other repository.

Is the link I've added to this readme to link to the Docker guide okay?

README.md
```diff
@@ -277,7 +277,10 @@ If you would like to change the name of the app or any other information you can
 
 #### Docker
 
-The recommended way to compile Scratch Everywhere! is with Docker. To compile with Docker all you need installed is Docker and Buildx.
+The recommended way to compile Scratch Everywhere! is with Docker. 
+
+We have written a step-by-step guide with pictures (for Windows) that explains the build process very simply, you can find it here: https://github.com/ScratchEverywhere/ScratchEverywhere.github.io/blob/main/docker.md
```
gradylink: This URL should be `https://ScratchEverywhere.github.io/docker`

Br0tcraft: fixed

PR #425, opened 2025-10-31
refactor(open project blocks):

Br0tcraft:
If you now start the project path with sd:/ in the `open() project` block, SE will no longer search in the Scratch Everywhere folder, but instead on the SD card.

Furthermore, it will first check if an extracted Scratch project is located in that folder path and it will then prioritize that one.

### DIFF
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -101,7 +101,7 @@ int main(int argc, char **argv) {
                 std::ofstream f(OS::getScratchFolderLocation() + filename);
                 f << buffer;
                 f.close();
-                Unzip::filePath = filename; //do i have to use OS::getScratchFolderLocation() + filename here?
+                Unzip::filePath = OS::getScratchFolderLocation() + filename;
                 Unzip::load(); // TODO: Error handling
             },
                                             &uploadComplete);
--- a/source/scratch/os.cpp
+++ b/source/scratch/os.cpp
@@ -160,7 +160,7 @@ bool Timer::hasElapsedAndRestart(int milliseconds) {
     return false;
 }
 
-std::string OS::getSDDrivePrefix() {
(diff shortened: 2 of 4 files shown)


source/scratch/os.cpp
```diff
@@ -160,32 +160,55 @@ bool Timer::hasElapsedAndRestart(int milliseconds) {
...
-    else return "fat:/scratch-ds/";
+    return isDSi() ? "sd:" : "fat:";
 #else
-    return "scratch-everywhere/";
+    return "";
 #endif
 }
 
+std::string OS::getScratchFolderLocation() {
+    std::string prefix = getSDDrivePrefix();
```
gradylink: Should be `const` or `constexpr`.

source/main.cpp
```diff
@@ -100,7 +101,7 @@ int main(int argc, char **argv) {
                 std::ofstream f(OS::getScratchFolderLocation() + filename);
                 f << buffer;
                 f.close();
-                Unzip::filePath = filename;
+                Unzip::filePath = filename; //do i have to use OS::getScratchFolderLocation() + filename here?
```
gradylink: Either that or remove the `OS::getScratchFolderLocation()` above. If you remove that, then you should also remove the line above it.

source/scratch/os.cpp
```diff
@@ -160,32 +160,55 @@ bool Timer::hasElapsedAndRestart(int milliseconds) {
     return false;
 }
 
-std::string OS::getScratchFolderLocation() {
+std::string OS::getSDDrivePrefix() {
```
gradylink: I feel like this should be called something else, since for some platforms it's not an SD Card. Maybe we should just call it `getFilesystemRootPrefix`?

gradylink (approved):

PR #427, opened 2025-11-01
Fix overly lenient string to double conversion

poipole807:
The previous conversion purely relied on `std::stod`, which converted the string to a double up until it couldn't. This was not correct compared to Scratch, as Scratch (JS) will not convert a string to a double if any of the characters are invalid.
Test cases: [Project.zip](https://github.com/user-attachments/files/23288109/Project.zip)

With this change, the math score on the performance tester went from 99.2% -> 97.9%.


source/scratch/math.cpp
```diff
@@ -35,6 +35,19 @@ int Math::color(int r, int g, int b, int a) {
...
+    while (std::isspace(static_cast<unsigned char>(str[pos])) && pos < str.length()) {
+        pos++;
+    }
+    if (str.length() != pos) {
+        throw std::invalid_argument("");
+    } else {
+        return conversion;
+    }
+}
+
```
gradylink: This function never seems to update the conversion variable.

poipole807: What do you suggest I do? Because I can't just move `std::stod(str, &pos)` to the return statement because it updates `pos` for the check

gradylink: Shouldn't you in the loop be doing `conversion += std::stod(str, &pos)` because right now you're only getting the first digit? Also, for `std::isspace(static_cast<unsigned char>(str[pos]))` just checks if a character is whitespace, shouldn't it check if it's a number?

gradylink (changes requested):
You need to run `clang-format` on your code, as it's using spaces instead of tabs for indents.

source/scratch/math.cpp
poipole807: I'm not quite sure if you're understanding what you're saying? The point of `str.length() != pos` is to check if `std::stod()` got cut short in its conversion, and if it did, throw an exception. However, if there is whitespace on the right-hand side of the string, the whitespace removal (which also happens in Scratch) gets mistaken for a conversion that got cut short. This is compensated with `pos++` by adding the whitespace that got truncated on the right edge of the string.

poipole807: Also, there's a much more serious bug in the function that I need to fix once clang installs

poipole807:
> You need to run `clang-format` on your code, as it's using spaces instead of tabs for indents.

How do I do that? I ran `clang -i source\scratch\math.cpp` and it does absolutely nothing

PR #432, opened 2025-11-03
PSP Port

NateXS:
This port will be using the PSP version of SDL2. Hopefully this won't be a PS4 port situation 😁😁😁

TO-DO List:

- [ ] ~~RapidJSON Migration (#352)~~ - (EDIT: NO LONGER NEEDED TO CONTINUE :fire: :fire:) (maybe something to do with us using CMake now?)

- [x] CmakeLists.txt edit for PSP
- [x] Get it to compile
- [x] Project loading
   - [ ] RomFS (upstream CMake bug)
   - [x] Main Menu
- [x] SDL2 things
   - [x] Rendering
   - [x] Input
   - [x] Audio
   - [x] Pen (texture is getting cut off though)
- [x] Dockerfile
- [ ] Cloud Variables (idk if this one is possible???)
- [x] PSP 1000 support (it'll be finee :grin:)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,10 +243,6 @@ endif()
 
 target_compile_definitions(scratch-everywhere PRIVATE __PC__)
 
-if(PSP)
-target_compile_definitions(scratch-everywhere PRIVATE __PSP__)
-endif()
-
 if(NOT SE_HEADLESS)
 	target_compile_definitions(scratch-everywhere PRIVATE SDL_BUILD)
 endif()


gradylink:
Cloud variables might be possible, but we would need to update their version of curl significantly

PwLDev:
I'm sure this won't be as hard as the PS4 because SDL2 has official support for the PSP. The reason why the PS4 port was challenging is due to abandoned toolchains and broken libs.

gradylink:
Btw, cmake romfs projects are currently broken fsr, I don't know why.

CMakeLists.txt
```diff
@@ -243,6 +243,10 @@ endif()
 
 target_compile_definitions(scratch-everywhere PRIVATE __PC__)
 
+if(PSP)
+target_compile_definitions(scratch-everywhere PRIVATE __PSP__)
+endif()
+
```
gradylink: Does pspdev not define this itself, or something similar at least that we can use?

gradylink (approved):

PwLDev:
that was fast

PR #436, opened 2025-11-04
Fix direction clamping and rounding

poipole807:
There was no clamping for the direction reporter, so the block could have a direction that was way larger than normal.

Test cases: [Project.zip](https://github.com/user-attachments/files/23323071/Project.zip)


gradylink (approved):
lgtm

PR #437, opened 2025-11-04
[Fix, SDL] Make streamed audio copy directly to memory where supported instead of creating a temp file

Dogo6647:
I noticed the audio loading system extracts sounds to disk every time a sound is played as streamed, so I figured that:

1. It would be much slower than copying directly to memory, and 
2. It would wear out flash storage devices like SD cards faster by making multiple write operations in a short time frame.

So I made some changes to the code to make it copy audio data directly to memory while keeping the legacy extract to temp file method as a fallback in case it fails (e.g. because of older SDL2).


Dogo6647:
oh that was quick

NateXS (approved):
Looks good to me!

PR #438, opened 2025-11-04
Fix change block when adding non-numerical value

poipole807:
Previously, when you changed a variable by a string, it would set the variable to that string instead of keeping the value intact.


gradylink (approved):
lgtm

PR #439, opened 2025-11-05
feat(sdl3): init

gradylink:
> [!NOTE]
> I am **not** adding Switch support to this PR, I will be adding it later in the CMake migration though.

# Checklist

- [x] Rendering
- [x] Audio
- [x] Pen
- [x] Input
	- [x] Keyboard
	- [x] Controller
	- [x] Mouse

### DIFF
--- a/docker/Dockerfile.macos
+++ b/docker/Dockerfile.macos
@@ -20,6 +20,7 @@ ARG ARCH=x86_64
 RUN mkdir build && cd build && \
 	${ARCH}-apple-darwin22.2-cmake \
 	-DCMAKE_OSX_DEPLOYMENT_TARGET=$MACOSX_DEPLOYMENT_TARGET \
+	-DSE_RENDERER=sdl2 \
 	-DSE_CLOUDVARS=OFF \
 	-DCMAKE_BUILD_TYPE=Release \
 	.. && \


source/scratch/blocks/pen.hpp
```diff
@@ -1,25 +1,6 @@
...
-#ifdef __3DS__
-#include <citro2d.h>
-#include <citro3d.h>
-
-extern C2D_Image penImage;
-extern C3D_RenderTarget *penRenderTarget;
-extern Tex3DS_SubTexture penSubtex;
-extern C3D_Tex *penTex;
-
-#define TEXTURE_OFFSET 16
```
NateXS, line 13: `TEXTURE_OFFSET` hasn't been moved to `3ds/render.cpp` like everything else here.

source/3ds/render.cpp
```diff
@@ -172,6 +177,78 @@ void Render::penMove(double x1, double y1, double x2, double y2, Sprite *sprite)
...
+    const float radius = thickness / 2.0f;
+
+    C2D_DrawCircleSolid(xSscaled, yScaled + TEXTURE_OFFSET, 0, radius, color);
+}
+
+void Render::penStamp(Sprite *sprite) {
+    const auto &imgFind = images.find(sprite->costumes[sprite->currentCostume].id);
+    if (imgFind == images.end()) {
+        Log::logWarning("Invalid Image for Stamp");
+        return BlockResult::CONTINUE;
```
NateXS: this should just be `return;` now

source/3ds/render.cpp
```diff
@@ -172,6 +177,78 @@ void Render::penMove(double x1, double y1, double x2, double y2, Sprite *sprite)
...
+    const ColorRGBA rgbColor = CSBT2RGBA(sprite->penData.color);
+    if (!Render::hasFrameBegan) {
+        if (!C3D_FrameBegin(C3D_FRAME_NONBLOCK)) C3D_FrameBegin(C3D_FRAME_SYNCDRAW);
+        Render::hasFrameBegan = true;
+    }
+    C2D_SceneBegin(penRenderTarget);
+    C3D_DepthTest(false, GPU_ALWAYS, GPU_WRITE_COLOR);
+
+    const int SCREEN_WIDTH = Render::getWidth();
+    const int SCREEN_HEIGHT = Render::getHeight();
```
NateXS: These should probably be renamed to something like `width` and `height` now, as the names are now conflicting with the #defines at the top of the file.

source/sdl3/keyboard.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "text.hpp"
+#include <SDL3/SDL.h>
+#include <string>
+
+#ifdef __SWITCH__
+#include <switch.h>
+#endif
+
+/**
+ * Uses SDL2 text input.
```
NateXS: SDL3* (nitpick i know 😭)

source/sdl3/stdout-wiiu.cpp
NateXS, line 1: Is this file needed? Wii U isn't using SDL3 as far as I know

source/scratch/blocks/pen.hpp
gradylink: I ended up just hard-coding it since it was only used once, if you want I can readd it.

PR #440, opened 2025-11-05
Text-to-Speech Extension for SE

Br0tcraft:
I’ve been working on a Text-to-Speech extension for the past two weeks.  It’s mostly done now, but there are a few limitations.

The API only lets you pick between male and female voices. Pitch and speed changes are handled locally (which I don’t think SE supports yet?).

I didn’t include the cat voice. I hope that’s fine.

The MP3s are saved to the SD card first and then loaded from there. If the file’s already there, it won’t be downloaded again.

I’m not entirely happy with the #ifdef-style checks for different platforms, but I didn’t come up with a better solution.

For Wii U, the API works fine without any certificates.
On 3DS, I had to include the certs in the romfs folder (romfs:/gfx/certs.pem).

The first four certs probably aren’t needed yet since they’re just from the standard Scratch domains. That might matter later once SE can directly download Scratch projects.
The newer ones are from the Scratch T2S API itself.

Right now, it probably only works on Wii U and 3DS. I tried it on Switch, but it still doesn’t seem to work there. It might also work on the PC builds, but I haven’t tested that.

Would be great if @NateXS could take a quick look to make sure the API usage is okay. TurboWarp also uses the Scratch API, so I don’t think there’ll be any issues, but since NateXS would be the one most affected, it’s probably best if he checks it.

(Also, I should have the translation extension ready in a few days!)


gfx/certs.pem
gradylink, line 1: Is there a reason we can't just use HTTP here? And if we do need to use HTTPS, we should probably be downloading these at build-time.

source/3ds/downloadManager.cpp
```diff
@@ -0,0 +1,195 @@
+#ifdef ENABLE_CLOUDVARS
```
gradylink: This shouldn't rely on cloud variables. You should either make it its own option or just make it compile against curl by default. The only reason cloud variables **are** an option is that they require curl to be built with specific options, which this does not, so I would recommend just making it build against curl by default.

source/sdl/render.cpp
```diff
@@ -199,6 +202,11 @@ bool Render::Init() {
     windowWidth = 1280;
     windowHeight = 720;
 #endif
+
+#ifdef ENABLE_CLOUDVARS
+    DownloadManager::init();
+#endif
```
gradylink: This doesn't need to be in a `#ifdef`.

source/3ds/render.cpp
```diff
@@ -60,6 +61,11 @@ bool Render::Init() {
     C2D_Prepare();
     gfxSet3D(true);
     C3D_DepthTest(false, GPU_ALWAYS, GPU_WRITE_COLOR);
+
+#ifdef ENABLE_CLOUDVARS
+    DownloadManager::init();
+#endif
```
gradylink: This doesn't need to be in a `#ifdef`.

gradylink (changes requested):

gfx/certs.pem
Br0tcraft: I tried a few things on the 3DS with the download, but somehow I never got it to work without a certificate. (It could also simply be my fault why it's not working)

source/3ds/downloadManager.cpp
Br0tcraft: Oh, I thought the cloud variables option was for deciding whether to use curl or not. Sorry.

Unfortunately, I don't quite understand how to do this so it doesn't conflict with the curl option for cloud variables. Could you help me with that?

source/sdl/render.cpp
Br0tcraft: Ah, right, I already had that there before I wrapped the entire file in an ifdef. I just forgot to delete it afterwards.

gfx/certs.pem
gradylink: Try using this code from Mist++: https://github.com/ScratchEverywhere/mistpp/blob/a1b8bfdab2452f29ce10f7074bccddf315090cf5/src/mist.cpp#L45-L46

source/3ds/downloadManager.cpp
gradylink: Actually it won't conflict at all, you can just add curl to the `pkg-config` call, and it will figure it out. They both just use the system curl package (except when using CMake) atm, it's just cloud variables only work (they still compile) if the system curl has been built with websockets enabled.

gfx/certs.pem
Br0tcraft: The two lines you linked to are already included in the downloader. However, you're right, no certificate is required there. (When downloading Scratch projects on the 3ds, I needed it in my other project because somehow it only worked in Citra, not on the 3DS itself.) With the Scratch Text-to-Speech API, though, it seems to work even without a certificate.

source/3ds/downloadManager.cpp
Br0tcraft: I don't know much about Makefiles/Dockerfiles. Could you perhaps do that for me?

gradylink: What you did should work fine.

Br0tcraft:
@gradylink Oh damn Thanks for looking over it. I seem to make a lot of (small) mistakes, sorry.

3DS and WiiU definitely work. I personally didn't get Switch to work. The PC builds should theoretically also work, I just haven't tested it yet. And I can't test PS4 myself or are there emulators that I can use for Homebrew PS4 (for later features)? 
That's why I haven't added them.

gradylink:
> @gradylink Oh damn Thanks for looking over it. I seem to make a lot of (small) mistakes, sorry.
> 
> 3DS and WiiU definitely work. I personally didn't get Switch to work. The PC builds should theoretically also work, I just haven't tested it yet. And I can't test PS4 myself or are there emulators that I can use for Homebrew PS4 (for later features)? That's why I haven't added them.

Yeah, ShadPS4 (the only PS4 emulator that actually works) does not work with SE! I could probably implement PC after this is merged, since it'll require a lot of stuff in the `CMakeLists.txt` (I tried to make the `CMakeLists.txt` as compatible as possible, but that also made it really complicated.)

PwLDev:
I could test on PS4 to see if it works. But I'm not sure if it will resolve because it was not resolving hostnames a few days ago while trying to implement cloudvars on the PS4.

Br0tcraft:
@NateXS could you then merge the pull request? If you have already tested it?

PR #442, opened 2025-11-06
feat(ui): change default font to Liberation Sans Narrow or Arimo

Dogo6647:
The current default UI font (Arial Narrow) is not free to use, so I'm thinking we should switch to either [Liberation Sans](https://en.wikipedia.org/wiki/Liberation_fonts), or [Arimo](https://fonts.google.com/specimen/Arimo), which look almost exactly the same and are meant to be metrically compatible with Arial.


gradylink (approved):

roccopm:
@gradylink This breaks #393. Do you guys want to merge this font first and then I'll make the same change for the speech classes before speech gets merged?

gradylink:
> @gradylink This breaks #393. Do you guys want to merge this font first and then I'll make the same change for the speech classes before speech gets merged?

It just depends on which gets merged first.

PR #446, opened 2025-11-07
Fix integer check for random block

poipole807:


source/scratch/value.cpp
```diff
@@ -214,6 +214,25 @@ bool Value::operator>(const Value &other) const {
...
+                return false;
+            }
+        }
+    } else if (isBoolean()) {
+        return true;
+    } else if (isString()) {
+        return asString().find('.') == std::string::npos;
+    }
+    return false;
+}
```
gradylink, line 229: We tend to try to avoid overly nesting (as it can make code harder to read.) All of these `else`s are entirely unnecessary because of the early escapes (returns).

```suggestion
bool Value::isScratchInt() {
  if (isDouble()) {
    if (std::isnan(asDouble())) return true;
    try {
      return std::stoi(asString()) == asDouble();
    } catch (...) {
      return false;
    }
  }
  if (isBoolean()) return true;
  if (isString()) return asString().find('.') == std::string::npos;
  return false;
}
```

Also, just curious, why not use `Math::parseNumber` here.

poipole807: `Math::parseNumber` is just unnecessarily slow here since the input is already guaranteed to be a double (although now that I think about it, a simple rounding check might be better here)

PR #448, opened 2025-11-08
NaN and comparative operator fixes

poipole807:
Both NaN and comparative operators were changed at the same time in order to avoid breaking something.

### DIFF
--- a/source/scratch/blocks/operator.cpp
+++ b/source/scratch/blocks/operator.cpp
@@ -149,9 +149,7 @@ Value OperatorBlocks::mathOp(Block &block, Sprite *sprite) {
 }
 
 Value OperatorBlocks::equals(Block &block, Sprite *sprite) {
-    Value value1 = Scratch::getInputValue(block, "OPERAND1", sprite);
-    Value value2 = Scratch::getInputValue(block, "OPERAND2", sprite);
-    return Value(value1 == value2);
+    return Value(Scratch::getInputValue(block, "OPERAND1", sprite) == Scratch::getInputValue(block, "OPERAND2", sprite));
 }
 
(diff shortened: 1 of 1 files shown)


source/scratch/value.cpp
```diff
@@ -42,17 +42,20 @@ bool Value::isColor() const {
     return std::holds_alternative<Color>(value);
 }
 
+bool Value::isNaN() const {
+    return std::holds_alternative<double>(value) && std::isnan(std::get<double>(value));
```
gradylink: To be consistent with the other functions, this should be:
```suggestion
    return isDouble() && std::isnan(asDouble());
```

source/scratch/blocks/operator.cpp
```diff
@@ -153,27 +150,9 @@ Value OperatorBlocks::mathOp(Block &block, Sprite *sprite) {
...
-        return Value(double1 == double2);
-    }
-    std::string string1 = value1.asString();
-    std::string string2 = value2.asString();
-    std::transform(string1.begin(), string1.end(), string1.begin(), ::tolower);
-    std::transform(string2.begin(), string2.end(), string2.begin(), ::tolower);
-    return Value(string1 == string2);
+    Value value1 = Scratch::getInputValue(block, "OPERAND1", sprite);
+    Value value2 = Scratch::getInputValue(block, "OPERAND2", sprite);
+    return Value(value1 == value2);
```
gradylink: You don't need intermediates in this case. I would do:
```suggestion
    return Value(Scratch::getInputValue(block, "OPERAND1", sprite) == Scratch::getInputValue(block, "OPERAND2", sprite));
```

source/scratch/value.cpp
poipole807: The whole point of the change is that I can't do `asDouble()`

PR #460, opened 2025-11-16
[SDL2] Specify project loader thread's stack size

Extrems:
This fixes a stack overflow on GameCube caught by https://github.com/extremscorner/libogc2/commit/e6b76febf47a28beb18391f46c415956f99266bb.

The Wii is also affected, but libogc's SDL2 port lacks the necessary plumbing for this change to have any effect. The issue is also silent there.


gradylink (approved):
lgtm

PR #461, opened 2025-11-16
Fix pen transparency-related issues

poipole807:


source/scratch/color.hpp
```diff
@@ -8,19 +8,17 @@ struct Color {
     float hue;
     float saturation;
     float brightness;
-
-    bool operator==(const Color &other) const {
-        return hue == other.hue && saturation == other.saturation && brightness == other.brightness;
-    }
+    float transparency;
```
gradylink, line 11: In the conversion functions you use `CSBO` for Color, Saturation, Brightness, Opacity which is the opposite of transparency. Opacity at 100% is solid, 0% fully transparent. Transparency is 100% fully transparent, 0% solid. Which is actually correct here

PR #474, opened 2025-11-23
docs: add separate control images

Starlii10:
Requirement of https://github.com/ScratchEverywhere/discord-bot/pull/2

I split the existing `controls.png` up and copied it to `controls/all.png` so that `.controls all` could exist.

Other than that, not much to say.


gradylink (changes requested):
You should remove the original `controls.png` and update the README to use `all.png` instead

PR #477, opened 2025-11-25
fix(nds): makefile

leap0x7b:
The DS Makefile had a typo in the BLOCKSDS variable and builds in build/ instead of the build/nds like other Makefiles. scratch-ds/ is also not included in the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,6 +71,7 @@ romfs/project.sb3_FILES/project.json
 Scratch-pc
 scratch-wiiu/
 scratch-wasm/
+scratch-ds/
 scratch-wii.zip
 
 # Stored Files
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -71,10 +71,14 @@ romfs/project.sb3_FILES/project.json
...
 scratch-wii.zip
 
 # Stored Files
 cloud-username.txt
 
 # macOS
 .DS_Store
+
+# fockerdile
+!docker/Dockerfile.nds
```
leap0x7b, line 84: Zed thinks Dockerfile.nds is ignored by Git even though it doesn't

gradylink (approved):
lgtm

PR #465, opened 2025-11-18
Optimized parsing logic

Lorna-1b:
Changes:
- Significantly improves performance on Math::parseNumber
- Uses std::optional instead of exceptions
- Math::parseNumber and Math::isNumber now take std::string_view
- values.cpp using those functions updated
- Moved repetitive parts of time.cpp to getLocalTime()

PR #478, opened 2025-11-26
fix(runtime): Replace std::rand() with new RNG to avoid modulo bias

Lorna-1b:
(Does not depend on PR [#465](https://github.com/ScratchEverywhere/ScratchEverywhere/pull/465))

Replaced `std::rand()` with the xoshiro128** PRNG (in the newly implemented `Random` class) for better quality and performance.

---

### Changes:
- Fixed modulo bias caused by using `rand() % range` to get a random range
- Moved `Math::generateRandomString()` to `Random::randomString()`
- Renamed `Math::removeQuotations` to `Math::removeDoubleQuotes` and updated to use `std::string_view` as input

### To-do:
- [ ] Add compile-time selection between 32-bit and 64-bit RNGs

### DIFF
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -87,9 +87,6 @@ int main(int argc, char **argv) {
         return 1;
     }
 
-    srand(time(NULL));
-
-    if (argc > 1) {
 #ifdef __EMSCRIPTEN__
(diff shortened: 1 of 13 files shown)


DevelopCMD:
why was this closed

PR #473, opened 2025-11-23
[WIP] Redo Entire Code Structure to be More Modular and General Cleanup

gradylink:
# Checklist

- [x] Use namespaces for blocks
- [x] Use macro magic to make adding new blocks easier
- [x] Improve folder structure
- [ ] Break up `CMakeLists.txt`
- [ ] Add shared make components
- [ ] Split platform-specific and rendering code
- [x] Split audio and rendering code
	- [ ] Separate audio initialization and render initialization
- [ ] Switch to namespace(s) for renderer and other static only classes and maybe singletons
- [ ] Refactor `BlockExecutor` to not be global/use an instance of itself in static functions
- [x] Cleanup deeply nested code and other outdated code/formatting mistakes
- [ ] Split `interpret.cpp` into `parser.cpp` and `runtime.cpp`
- [x] Add missing `const`/`constexpr`
- [ ] Cleanup logs
	- [ ] Add missing NOP blocks
	- [ ] Clear file on launch or use multiple log files (setting?)
	- [x] Update `std::cerr`/`std::cout` uses to use `Log` class
- [ ] Runtime Renderer Switching (well, not after the app has launched, but I mean be able to change after it's installed/built)
- [ ] Add a better way to add custom blocks
- [ ] Add shared sprite drawing function
- [ ] Don't use `SoundPlayer` at all if audio is disabled
- [ ] Update Dockerfiles and nightly actions with new changes
- [ ] Add options to the Dockerfiles
- [ ] Make sure all the platforms and renderers (all possible combinations) compile.
- [ ] Abstract unzipping to allow for multiple libraries

### PLEASE GIVE FEEDBACK, THIS PR IS LIKELY GOING TO BE VERY OPINIONATED!!!

PR #480, opened 2025-11-28
feat(3ds,wiiu): Support for T2S blocks

Br0tcraft:
GitHub had problems merging (or maybe I was just too dense). Anyway, it should hopefully work now. Support for theoretically every SDL version as well as 3DS. (With 3DS, though, I'm currently having the problem that no sound is playing at all, even with Docker.)

### DIFF
--- a/source/scratch/os.hpp
+++ b/source/scratch/os.hpp
@@ -97,7 +97,8 @@ class OS {
      * De-Initializes the internet.
      */
     static void deInitWifi();
-  
+
+    /**
      * Create a directory.
      * @param path The path of the directory to create.
      */


NateXS:
for the 3DS, im getting this warning both in the Dockerfile and locally
```
Package curl was not found in the pkg-config search path.
Perhaps you should add the directory containing `curl.pc'
to the PKG_CONFIG_PATH environment variable
Package 'curl', required by 'virtual:world', not found
```
this might be why there's nothing playing on the 3DS side of things

gradylink:
> for the 3DS, im getting this warning both in the Dockerfile and locally
> 
> ```
> Package curl was not found in the pkg-config search path.
> Perhaps you should add the directory containing `curl.pc'
> to the PKG_CONFIG_PATH environment variable
> Package 'curl', required by 'virtual:world', not found
> ```
> 
> this might be why there's nothing playing on the 3DS side of things

Yep, the correct name is `libcurl` not `curl`.

Br0tcraft:
Thanks! I think it should work now.

gradylink:
@Br0tcraft still an issue with Wii U

Br0tcraft:
@gradylink Really? What exactly is the problem with the Wii U? I can compile the source code without any problems using Docker, and the sound is downloaded to the cache folder and played back (but I used Cemu for testing, could that be the reason?).

gradylink:
> @gradylink Really? What exactly is the problem with the Wii U? I can compile the source code without any problems using Docker, and the sound is downloaded to the cache folder and played back (but I used Cemu for testing, could that be the reason?).

@Br0tcraft Wii U also uses `libcurl` for the `pkg-config` name, if you were compiling with Docker it would be adding the correct package from Mist++, but if you're compiling without cloud variables then it will fail to find libcurl.

Br0tcraft:
@gradylink Oh sorry, it should be changed now.

NateXS:
Tested again on 3DS, and now im getting warning
```
Warning: Failed to load audio file: sdmc:/3ds/scratch-everywhere/cache/t2s_temp_3653975854.mp3 - SDL_mixer Error: Audio data is in unknown/unsupported/corrupt format
```
and the sound file in my cache folder is 0 bytes in size.

EDIT: ok i noticed a more telling errror
```
Download failed: Couldn't resolve host name
```

NateXS:
Ended up fixing it, it was because it wasn't working if cloud variables wasn't enabled

gradylink:
I just realized how much this is going to fuck up #473

PR #481, opened 2025-11-28
feat: remove dependency on std::filesystem and make cmakerc an option

leap0x7b:


CMakeLists.txt
```diff
@@ -70,6 +70,10 @@ if(NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/cmake/CPM.cmake")
 endif()
 include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/CPM.cmake")
 
+if(NOT NINTENDO_SWITCH AND NOT VITA)
+	option(SE_CMAKERC "Embed romfs to executable with CMakeRC instead of pulling from the working directory" ON)
```
leap0x7b: yeah i don't know how to phrase this

gradylink: Instead of "romfs" you could say "assets"?

source/scratch/menus/controlsMenu.cpp
```diff
@@ -218,8 +218,8 @@ void ControlsMenu::applyControls() {
 
     // Make sure parent directories exist
     try {
-        std::filesystem::create_directories(std::filesystem::path(filePath).parent_path());
-    } catch (const std::filesystem::filesystem_error &e) {
+        OS::createDirectory(OS::parentPath(filePath));
+    } catch (const std::exception &e) {
```
gradylink: Shouldn't we make our own `OS::fs_exception` here in order to differentiate it from other exceptions that may occur (IDK what those would be, but it's still good practice.)

leap0x7b: i have no idea how to do that but i could try

gradylink (approved):

PR #482, opened 2025-11-29
Add makey makey blocks

poipole807:
Adds Makey Makey blocks and makes other key-sensing related changes.

Test cases: [Project.zip](https://github.com/user-attachments/files/23831887/Project.zip)


source/scratch/input.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    std::transform(key.begin(), key.end(), key.begin(), ::tolower);
+    if (key.length() > 0) {
+        key = key.substr(0, 1);
+    }
+
+    return key;
+}
+
+void Input::executeKeyHats() {
```
gradylink: This should be in the `BlockExecutor` class (so should `doSpriteClicking` so if you could move that while you're at it, that'd be great).

gradylink (changes requested):

NateXS:
This PR isn't compiling on 3DS or NDS, there's a bunch of undefined references when linking

poipole807:
> This PR isn't compiling on 3DS or NDS, there's a bunch of undefined references when linking

Should be fixed

PR #485, opened 2025-11-30
fix(runtime): Avoid overflows when using `Value::asString()` with doubles

a-church:
Uses `std::snprintf()` to automatically format doubles. A more modern option might be `std::format`, but that won't work on older platforms that don't have a modern implementation of `std` (specifically C++20).

### DIFF
--- a/source/scratch/value.cpp
+++ b/source/scratch/value.cpp
@@ -64,10 +64,16 @@ std::string Value::asString() const {
         return std::to_string(std::get<int>(value));
     } else if (isDouble()) {
         double doubleValue = std::get<double>(value);
-        int bufferSize = std::snprintf(nullptr, 0, "%g", doubleValue);
-        std::vector<char> buf(bufferSize + 1);
-        std::snprintf(buf.data(), buf.size(), "%g", doubleValue);
-        return std::string(buf.data(), buf.data() + bufferSize);
+        // NaN values MUST return "NaN", and Inf values MUST return either
+        // "-Infinity" or "Infinity" string constants
+        if (std::isnan(doubleValue)) return "NaN"; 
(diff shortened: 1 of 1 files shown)


gradylink:
Does this new system still properly handle NaN, Infinity, and -Infinity.

garrickch:
> Does this new system still properly handle NaN, Infinity, and -Infinity.

It does [(cppreference)](https://en.cppreference.com/w/c/io/fprintf).

```c++
#include <string>
#include <vector>
#include <cmath>
#include <iostream>

std::string parse(double doubleValue) {
        int bufferSize = std::snprintf(nullptr, 0, "%g", doubleValue);
        std::vector<char> buf(bufferSize + 1);
        std::snprintf(buf.data(), buf.size(), "%g", doubleValue);
        return std::string(buf.data(), buf.data() + bufferSize);
}

int main(int argc, char* argv[]) {
        std::cout << parse(NAN) << "\n"; // nan
        std::cout << parse(INFINITY) << "\n"; // inf
        std::cout << parse(-INFINITY) << "\n"; // -inf
}
```

NateXS (approved):

gradylink:
@a-church It needs to be specifically `NaN`, `Infinity`, and `-Infinity`, not `nan`, `inf`, and `-inf`; that's why those checks were there in the first place.

garrickch:
Accidentally closed lol :stuck_out_tongue: 
Last commits re-added the special NaN and Inf formatting. Currently checking to make sure that behavior of truncation isn't completely borked because I noticed some weird stuff in testing.

gradylink:
This has been replaced by poipole's fork of ryuJS

PR #486, opened 2025-12-01
fix(windows): build issue on msvc

leap0x7b:
also adds windows resources and partial hidpi support on sdl2


leap0x7b:
the changes i made for msvc should also work for mingw as well but i haven't tested it out yet

PR #489, opened 2025-12-03
feat(monitor): improved appearance 

leap0x7b:
Made the monitor look and behave closer to the official Scratch runtime. I planned to add slider and list support though I can't figure out how. I also made some miscellaneous changes to SDL1 and Windows.

Here's a screenshot of it, running [Itch's monitor test](https://github.com/mgurga/itch/blob/master/test/MonitorTest/monitortest.sb3):
Jepretan layar_20251203_144946

### DIFF
--- a/source/scratch/blockExecutor.cpp
+++ b/source/scratch/blockExecutor.cpp
@@ -673,38 +673,21 @@ void BlockExecutor::updateMonitors() {
                         parsedField.value = Math::removeQuotations(paramValue);
                         (*newBlock.parsedFields)[paramName] = parsedField;
                     }
-                    if (var.opcode == "motion_xposition")
-                        var.displayName = var.spriteName + ": x position";
-                    else if (var.opcode == "motion_yposition")
-                        var.displayName = var.spriteName + ": y position";
-                    else if (var.opcode == "motion_direction")
-                        var.displayName = var.spriteName + ": direction";
-                    else if (var.opcode == "sound_volume")
-                        var.displayName = var.spriteName + ": volume";
-                    else if (var.opcode == "looks_size")
-                        var.displayName = var.spriteName + ": size";
-                    else if (var.opcode == "looks_costumenumbername")
+                    if (var.opcode == "looks_costumenumbername")
                         var.displayName = var.spriteName + ": costume " + Scratch::getFieldValue(newBlock, "NUMBER_NAME");
                     else if (var.opcode == "looks_backdropnumbername")
                         var.displayName = "backdrop " + Scratch::getFieldValue(newBlock, "NUMBER_NAME");
-                    else if (var.opcode == "sensing_timer")
-                        var.displayName = "timer";
-                    else if (var.opcode == "sensing_username")
-                        var.displayName = "username";
-                    else if (var.opcode == "sensing_loudness")
-                        var.displayName = "loudness";
-                    else if (var.opcode == "sensing_answer")
-                        var.displayName = "answer";
-                    else if (var.opcode == "sensing_current") {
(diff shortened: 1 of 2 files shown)


leap0x7b:
I only tested this with SDL1 and SDL2. SDL3 is probably identical with the other SDL versions though I don't know how would it look like on 3DS and DS.

source/scratch/blockExecutor.cpp
```diff
@@ -604,82 +605,113 @@ void BlockExecutor::setVariableValue(const std::string &variableId, const Value
...
+                    else if (var.opcode == "sensing_current") {
+                        std::string inputValue = Scratch::getFieldValue(newBlock, "CURRENTMENU");
+                        if (inputValue == "YEAR") var.displayName = "year";
+                        if (inputValue == "MONTH") var.displayName = "month";
+                        if (inputValue == "DATE") var.displayName = "date";
+                        if (inputValue == "DAYOFWEEK") var.displayName = "day of week";
+                        if (inputValue == "HOUR") var.displayName = "hour";
+                        if (inputValue == "MINUTE") var.displayName = "minute";
+                        if (inputValue == "SECOND") var.displayName = "second";
+                    } else var.displayName = var.opcode;
```
gradylink: Why not use a `constexpr` (if you can't do `constexpr` you can still do `const`) map with a lookup here? Or, at least for more of them, you could do a `constexpr std::unordered_map<std::string, std::function<std::string(Block &newBlock)>> myMap = { /* My Items */  }` then do `var.displayName = myMap[var.opcode]` (with checks for unknown opcodes around ofc.)

source/scratch/render.hpp
```diff
@@ -195,6 +195,23 @@ class Render {
...
+        else if (opcode.substr(0, 8) == "sensing_")
+            return {.r = 92, .g = 177, .b = 214, .a = 255};
+        else if (opcode.substr(0, 7) == "motion_")
+            return {.r = 76, .g = 151, .b = 255, .a = 255};
+        else if (opcode.substr(0, 6) == "looks_")
+            return {.r = 153, .g = 102, .b = 255, .a = 255};
+        else if (opcode.substr(0, 6) == "sound_")
+            return {.r = 207, .g = 99, .b = 207, .a = 255};
+        else return {.r = 255, .g = 140, .b = 26, .a = 255};
+    }
```
gradylink, line 213: Could also be a map, you could detect up to the first `_` then compare.

gradylink (commented):
More of a question as to why not choose that, since it seems cleaner in my eyes at least.

gradylink:
> I only tested this with SDL1 and SDL2. SDL3 is probably identical with the other SDL versions though I don't know how would it look like on 3DS and DS.

Hey, they're compiling at least.

source/scratch/blockExecutor.cpp
leap0x7b: I have no idea how to do that but I guess I could try

source/scratch/render.hpp
leap0x7b: no idea how to do that so i just kept it if-else statements

NateXS:
3DS:
Screenshot from 2025-12-03 06-34-18

DS:
Screenshot from 2025-12-03 06-38-09

The 3DS position looks a bit off because i do `yPos -= getSize()[1] / 2;` when rendering text, it fixes some text and breaks others for some reason? it might have to do with different fonts, so I'll fix that after this PR

DS Text doesn't support scale, not because it's hard to implement but because if the text is any smaller it becomes way less readable. IDK though, I'll probably make it support scale anyway

PR looks good otherwise though

PR #493, opened 2025-12-10
feat(collision): more accurate collision system

gradylink:
# Checklist

- [x] Cleanup Existing Code
- [x] Add necessary methods to `Image` class
	- [ ] NDS (even though it will still use rect collision)
	- [ ] 3DS
	- [ ] SDL1
	- [x] SDL2
	- [ ] SDL3
- [ ] Migrate code from https://github.com/gradylink/se-hitbox-demo
- [ ] Rotation Support
- [ ] Add settings

### DIFF
--- a/source/scratch/interpret.cpp
+++ b/source/scratch/interpret.cpp
@@ -470,60 +470,92 @@ bool isColliding(std::string collisionType, Sprite *currentSprite, Sprite *targe
             }
         }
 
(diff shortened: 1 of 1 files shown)


gradylink:
Sorry for delaying this again, but #546 is adding some features that will be extremely helpful for this PR, so I'm going to wait for that to be added then make a new PR.

PR #494, opened 2025-12-12
Fix case-sensitive effect name comparison

kalomaze:
Effect names in project.json can be lowercase in some contexts (e.g., "ghost"), but the code compared against uppercase ("GHOST"). Added std::transform to normalize effect names to uppercase before comparison.


gradylink:
It might be worth checking if this affects other blocks that use systems like this, I'm thinking maybe the pen blocks?

PR #495, opened 2025-12-12
[Audio] Fix IMA ADPCM loading and channel collision

kalomaze:
- Implement in-memory ADPCM to PCM decoder
- Fix channel collision: verify chunk before halting

SDL_mixer fails to decode ADPCM from RWops. Afaic this is the least bad fix given constraints: no temp files (disk I/O), no new deps, no file bloat from pre-converting audio (also disk I/O), and an upstream SDL fix(?) is out of scope.

Repro: https://turbowarp.org/258765038 (music wouldn't play)

### DIFF
--- a/source/sdl2/audio/audio.cpp
+++ b/source/sdl2/audio/audio.cpp
@@ -19,6 +19,196 @@ std::unordered_map<std::string, std::unique_ptr<SDL_Audio>> SDL_Sounds;
 std::string currentStreamedSound = "";
 static bool isInit = false;
 
+// IMA ADPCM step table
+static const int ima_step_table[89] = {
+    7, 8, 9, 10, 11, 12, 13, 14, 16, 17,
+    19, 21, 23, 25, 28, 31, 34, 37, 41, 45,
+    50, 55, 60, 66, 73, 80, 88, 97, 107, 118,
+    130, 143, 157, 173, 190, 209, 230, 253, 279, 307,
+    337, 371, 408, 449, 494, 544, 598, 658, 724, 796,
+    876, 963, 1060, 1166, 1282, 1411, 1552, 1707, 1878, 2066,
+    2272, 2499, 2749, 3024, 3327, 3660, 4026, 4428, 4871, 5358,
+    5894, 6484, 7132, 7845, 8630, 9493, 10442, 11487, 12635, 13899,
+    15289, 16818, 18500, 20350, 22385, 24623, 27086, 29794, 32767};
+
+// IMA ADPCM index table
+static const int ima_index_table[16] = {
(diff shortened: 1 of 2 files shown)


gradylink:
I personally think upstreaming a fix to SDL2_Mixer makes more sense (if they don't already have one in their dev branch), we've upstreamed many fixes before, also make sure you're it's not just an issue with an outdated version of SDL2_Mixer, are you using the PC version compiled from source with non-system libs?

kalomaze:
> I personally think upstreaming a fix to SDL2_Mixer makes more sense (if they don't already have one in their dev branch)

IC, I can look into that

> also make sure you're it's not just an issue with an outdated version of SDL2_Mixer, are you using the PC version compiled from source with non-system libs?

The song fails to play on the Wii nightly, I don't believe it's an issue localized to my build. Also now that you're mentioning the platform, I'm compiling and testing my fix on PowerPC (Wii); seems the endianness is clearly wrong, about to commit a fix or two for this. (Also I think stereo audio support isn't necessary for Scratch or even functional lol whoops)

gradylink:
> The song fails to play on the Wii nightly, I don't believe it's an issue localized to my build. Also now that you're mentioning the platform, I'm compiling and testing my fix on PowerPC (Wii); seems the endianness is clearly wrong, about to commit a fix or two for this. (Also I think stereo audio support isn't necessary for Scratch or even functional lol whoops)

If the issue only happens on Wii, we should upstream a fix to https://github.com/devkitpro/pacman-packages, first checking if updating their version of SDL2_mixer fixes the issue and if that doesn't work using a patch file on the SDL2_mixer wii package

kalomaze:
> If the issue only happens on Wii, we should upstream a fix to https://github.com/devkitpro/pacman-packages, first checking if updating their version of SDL2_mixer fixes the issue and if that doesn't work using a patch file on the SDL2_mixer wii package

It's not only on Wii, there was silence on my Mac build of `main` as well (`Unknown WAVE data format`)
Latest commit compiles and plays the music properly on both Dolphin when I load the .dol, & on the native Mac build w/o the prior garble from wrong endianness.

gradylink:
> It's not only on Wii, there was silence on my Mac build of main as well (Unknown WAVE data format)

what command did you use to configure on Mac.

kalomaze:
> > It's not only on Wii, there was silence on my Mac build of main as well (Unknown WAVE data format)
> 
> what command did you use to configure on Mac.

`cmake .. -DSE_RENDERER=sdl2`

And then to build:
`make -j8`

Looks like I am on SDL2_mixer 2.8.1 (latest).
Worth noting the problem is reproducible in the nightlies for Wii.

gradylink:
This should no longer be necessary with our custom audio mixer.

PR #496, opened 2025-12-13
[Runtime] Fix case-sensitive broadcast matching

kalomaze:
- Normalize broadcast names to lowercase before comparison

Scratch broadcast matching is case-insensitive, SE was doing exact match

Repro: https://turbowarp.org/258765038 (battle menu softlock)


kalomaze:
Not sure if this is a problem anywhere else as suspected here
https://github.com/ScratchEverywhere/ScratchEverywhere/pull/494#issuecomment-3648471658

PR #497, opened 2025-12-13
[Runtime] Preserve project order for sounds and block chains

kalomaze:
- Add soundOrder vector for index-based sound lookup
- Add blockChainOrder vector for deterministic script execution

std::map iterates alphabetically, unordered_map is non-deterministic. Caused wrong sounds on index lookup and script execution order varying per launch (flickering sprites for `forever` loops in the case of the project I was testing)

Repro: https://turbowarp.org/258765038 (flickering)

### DIFF
--- a/source/scratch/interpret.cpp
+++ b/source/scratch/interpret.cpp
@@ -821,8 +821,8 @@ void loadSprites(const nlohmann::json &json) {
             newSound.dataFormat = data["dataFormat"];
             newSound.sampleRate = data["rate"];
             newSound.sampleCount = data["sampleCount"];
+            newSound.insertionOrder = newSprite->sounds.size();
             newSprite->sounds[newSound.name] = newSound;
-            newSprite->soundOrder.emplace_back(newSound.name);
         }
 
         // set Costumes
@@ -1118,8 +1118,8 @@ void loadSprites(const nlohmann::json &json) {
             std::string outID;
             BlockChain chain;
             chain.blockChain = getBlockChain(block.id, &outID);
+            chain.insertionOrder = currentSprite->blockChains.size();
             currentSprite->blockChains[outID] = chain;
-            currentSprite->blockChainOrder.emplace_back(outID);
             block.blockChainID = outID;
 
             for (auto &chainBlock : chain.blockChain) {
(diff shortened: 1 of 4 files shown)


source/scratch/sprite.hpp
```diff
@@ -209,11 +209,13 @@ class Sprite {
     std::map<std::string, Block> blocks;
     std::unordered_map<std::string, List> lists;
     std::map<std::string, Sound> sounds;
+    std::vector<std::string> soundOrder; // Preserves project order for index-based lookup
```
gradylink: The order is preserved in an `std::map` so you shouldn't need a separate vector here, in other places the issue is that you need to go backwards so that might also be the case here.

source/scratch/sprite.hpp
```diff
@@ -209,11 +209,13 @@ class Sprite {
...
     std::vector<Costume> costumes;
     std::unordered_map<std::string, Comment> comments;
     std::unordered_map<std::string, Broadcast> broadcasts;
     std::unordered_map<std::string, CustomBlock> customBlocks;
     std::unordered_map<std::string, BlockChain> blockChains;
+    std::vector<std::string> blockChainOrder; // Preserves project order for deterministic execution
```
gradylink: To avoid duplicated data, you should just use an `std::map` here to preserve order.

source/scratch/sprite.hpp
kalomaze: `std::map` doesn't appear to replicate the json order for the particular project I was testing here.
It didn't seem like there was a clear or consistent alphabetical pattern that I could exploit via heuristic.

`['hurt noise', 'select', 'move selection']` is the canon project order; this doesn't match any alphabetical variant(?)

source/scratch/interpret.cpp
```diff
@@ -822,6 +822,7 @@ void loadSprites(const nlohmann::json &json) {
             newSound.sampleRate = data["rate"];
             newSound.sampleCount = data["sampleCount"];
             newSprite->sounds[newSound.name] = newSound;
+            newSprite->soundOrder.push_back(newSound.name);
```
gradylink: To fix the issue you mentioned in the other comment, replace this `sounds` value set stuff to use `emplace_back`.

PR #501, opened 2025-12-16
fix(input): mouse support on sdl2 and 3

leap0x7b:
This is actually a 2-in-1 PR lol. I originally wanted this to be about adding support for running projects through CLI arguments but then I encountered broken mouse input on SDL2 and 3 while testing it with Hopper Heroes so I just decided to make this about fixing mouse support just because.


gradylink:
I think I'll just wait for checks to finish, then I'll merge this.

PR #509, opened 2025-12-25
fix(cloudvars): use system curl on haiku

leap0x7b:


gradylink:
Do cloud variables work with system curl?

leap0x7b:
haven't tested it out yet though it should work, do you have an example sb3 file?

leap0x7b:
https://github.com/user-attachments/assets/9d025d78-9891-49ab-b2fc-05642e083a55

can confirm its working (pardon for the lag, this is in a vm so)

PR #384, opened 2025-10-04
more splashText quotes, and UserName var

DevelopCMD:
self explanitory ig

PR #511, opened 2025-12-26
Splash text {UserName} var

roccopm:
This is a re-implementation of the seemingly abandoned PR #384.
As well as adding this feature, I have moved the implementations of the unzip functions from the header to the .cpp file (as this is standard practice and should reduce compile times).

To test, I replaced all the splash text temporarily with one line: `Wazzup {UserName} on {PlatformName}` to ensure that both vars still get replaced as expected:
image

### DIFF
--- a/gfx/menu/splashText.txt
+++ b/gfx/menu/splashText.txt
@@ -28,7 +28,7 @@ uiText.titleScreen.splashText14
 FUNNY AND/OR CLEVER SPLASH TEXT
 You'll love it!
 Magic!
-Try it out!!!!
+Try it out, {UserName}!!!!
 I'm itchy everywhere, what should I- oh, right...
 Join the Discord Server!
 Contribute on GitHub!
@@ -59,7 +59,7 @@ Woah.
 Are wii gonna have a problem?
 Beta!
 Deja vu!
-Let's play!
+Let's play, {UserName}!
 Woo, TurboWarp!
 Woo, ScratchBox!
 Woo, softwaregore!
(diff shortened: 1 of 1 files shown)


source/runtime/unzip.cpp
```diff
@@ -207,3 +224,316 @@ bool Unzip::load() {
...
+    while ((pos = splash.find(platformPlaceholder, pos)) != std::string::npos) {
+        splash.replace(pos, platformPlaceholder.size(), platform);
+        pos += platform.size(); // move past replacement
+    }
+
+    pos = 0;
+    while ((pos = splash.find(usernamePlaceholder, pos)) != std::string::npos) {
+        splash.replace(pos, usernamePlaceholder.size(), username);
+        pos += username.size(); // move past replacement
+    }
```
roccopm, line 381: This is the only part of the implementation in this file that I changed. Everything else is just moved over from the .hpp

source/main.cpp
```diff
@@ -25,7 +25,12 @@ static void exitApp() {
...
 
 static bool initApp() {
-    return Render::Init();
+    if (!Render::Init()) {
+        return false;
+    }
+
+    loadUsernameFromSettings();
+    return true;
 }
```
roccopm, line 34: Loading username on app init so that the saved value is accessible in the menu. When calling Input::getUsername in `unzip.cpp`, it was returning Player because settings.json hadn't been loaded yet

roccopm:
Final test after resolving merge conflicts:
image

Testing on a fresh project (no Settings.json):
image

PR #507, opened 2025-12-22
feat(opengl): init

leap0x7b:
A basic OpenGL (ES) 2.0 renderer which can use other windowing libraries (GLFW, SDL2, SDL3).  May add libretro support to this as per the platform support roadmap in the future.

PR #512, opened 2025-12-27
feat(opengl): init

leap0x7b:
continuation from #507, really disappointed the restructure still doesn't separate the windowing from the renderer but oh well

also adds a generic pthread and windows implementation for the unzip thread

### DIFF
--- a/source/renderers/opengl/render.cpp
+++ b/source/renderers/opengl/render.cpp
@@ -170,8 +170,6 @@ static void preparePenDrawing() {
 
     glClearColor(0.0f, 0.0f, 0.0f, 0.0f);
     glClear(GL_COLOR_BUFFER_BIT);
-    glDisable(GL_SCISSOR_TEST); // we want to draw lines starting from outside
-    glEnable(GL_SCISSOR_TEST);  // wait actually we want to clip to texture bounds
 
     glMatrixMode(GL_PROJECTION);
     glPushMatrix();


gradylink:
> doesn't separate the windowing from the renderer but oh well

Reminder that PR was merged well before it was done, the rest of the changes will be added later.

source/renderers/opengl/sdl2/input.cpp
gradylink, line 1: Isn't this identical to the version in `renderers/sdl2/input.cpp`, couldn't we just use that file with an `#ifdef` for `render.hpp` or `../opengl/sdl2/window.hpp`. Or you could implement splitting the renderer from the window (which would be preferred).

leap0x7b:
> > doesn't separate the windowing from the renderer but oh well
> 
> Reminder that PR was merged well before it was done, the rest of the changes will be added later.

fair enough

source/renderers/opengl/sdl2/input.cpp
leap0x7b: yea its identical, made it so that it uses those files instead, i'll just do the renderer-window split in a separate pr instead since i'm really burned out rn

source/renderers/opengl/sdl2/window.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+#include "../window.hpp"
+#include <SDL2/SDL.h>
+
+class Sdl2Window : public Window {
```
gradylink: Preferably, this should match the same naming scheme as `TextObjectSDL2`.

source/renderers/sdl2/stdout-wiiu.cpp
gradylink, line 1: Could you revert this file to its state before it's not meant to be formatted (I'll update the `.clang-format-ignore` in main, mb.)

source/runtime/unzip.cpp
```diff
@@ -210,6 +226,38 @@ bool Unzip::load() {
...
+        Loading loading;
+        loading.init();
+
+        while (!Unzip::threadFinished) {
+            loading.render();
+        }
+        pthread_join(thread, NULL);
+        loading.cleanup();
+    } else Unzip::openScratchProject(NULL);
+
```
gradylink: Is there a reason we can't just use `std::thread` here?

leap0x7b: i forgor that exists

source/renderers/opengl/render.cpp
```diff
@@ -0,0 +1,586 @@
...
+    glPushAttrib(GL_VIEWPORT_BIT | GL_COLOR_BUFFER_BIT | GL_ENABLE_BIT | GL_TEXTURE_BIT);
+
+    glViewport(0, 0, penWidth, penHeight);
+    glScissor(0, 0, penWidth, penHeight);
+    glEnable(GL_SCISSOR_TEST);
+
+    glClearColor(0.0f, 0.0f, 0.0f, 0.0f);
+    glClear(GL_COLOR_BUFFER_BIT);
+    glDisable(GL_SCISSOR_TEST); // we want to draw lines starting from outside
+    glEnable(GL_SCISSOR_TEST);  // wait actually we want to clip to texture bounds
```
NateXS: what is the point of this :sob:

NateXS: this code does nothing :sob:

NateXS (changes requested):
Other than that, I tested it and it works good!

NateXS (approved):

PR #516, opened 2025-12-28
refactor: separate windowing from renderer

leap0x7b:
also added opengl support for sdl1, i only tested this on opengl and sdl i'm def going get cooked by the reviews and requested changes

### DIFF
--- a/source/windowing/3ds/input.cpp
+++ b/source/windowing/3ds/input.cpp
@@ -44,7 +44,7 @@ void Input::getInput() {
     inputButtons.clear();
     mousePointer.isPressed = false;
     mousePointer.isMoving = false;
-    if (globalWindow) globalWindow->pollEvents();
+    hidScanInput();
     u32 kDown = hidKeysHeld();
 
     hidTouchRead(&touch);


leap0x7b:
yeah i expected ds and 3ds to fail

leap0x7b:
also fixed some build errors on sdl1 made after the restructure, it's still broken though and i still have no idea why, might try to fix it in a future pr
image

leap0x7b:
istg if this ends up like the opengl pr

source/windowing/sdl2/input.cpp
leap0x7b, line 1: oops

leap0x7b: its fixed on the next commit anyway who cares

leap0x7b:
why only 3ds, ds, and psp are enabled for pull requests anyway? is there something that prevents all the platforms to not build on prs for whatever reason cus there doesn't seem to be one afaik, would be really useful so that i don't have to setup a focker for every platform and not worry about running out of space on my 256gb ssd

gradylink:
3DS, DS, and PSP cover the three major renderers, SDL2, GL2D, and Citro2D.

CMakeLists.txt
```diff
@@ -72,16 +72,65 @@ if(NOT SE_RENDERER IN_LIST SE_RENDERER_VALID_OPTIONS)
...
+			"headless"
+			CACHE STRING "Select the desired windowing backend."
+		)
+	endif()
+	if(NOT SE_WINDOWING IN_LIST SE_WINDOWING_VALID_OPTIONS)
+	    message(
+	        FATAL_ERROR
+	        "Invalid value for SE_WINDOWING: '${SE_WINDOWING}'. Must be one of: ${SE_WINDOWING_VALID_OPTIONS}"
+	    )
+	endif()
```
gradylink: The error check should be moved before the "if else" chain, and this entire block should be moved out of the OpenGL specific renderer in the case of future renderers like Vulkan being added later, then the other renderers should only set the valid options.

gradylink (changes requested):
To match `renderers`, I think the `window` folder should be renamed to `windows`.

roccopm:
> also fixed some build errors on sdl1 made after the restructure, it's still broken though and i still have no idea why, might try to fix it in a future pr image

I fixed this rendering issue in the say/think PR, it was a super simple change:
image

`spriteSurface` gets freed before this check so it always returned true. `spriteTexture` is what's actually getting rendered so I believe this is a safe change to make to just check that on its own.
The reason the images would sometimes display is _only_ for SVG images as for that execution path, spriteSurface doesn't get freed but it can contain garbled data because it's never initialised (doesn't need to be for SVG, so it would come back as _not_ null even though it doesn't contain anything useful).

I think I'm going to take this out of say/think and open a separate PR for this tomorrow after work, but don't have time tonight so feel free to do that yourself / implement the fix into this PR.

leap0x7b:
ok i get it it's a typo too long didn't read kthxbi

CMakeLists.txt
leap0x7b: the error check for renderer and audio engine appeared after the if else chain

gradylink: @leap0x7b you didn't implement this, yet you marked it as resolved.

leap0x7b: wdym

leap0x7b: i just moved the block to else because it doesn't really make sense to use an sdl3 renderer with sdl2 windowing lol (and i don't think it would even work in the first place)

leap0x7b: also why do i need to move the error check for windowing to before the chain if the renderer and audio engine has it after that, do i need to move the error check for those as well or what?

gradylink: > i just moved the block to else because it doesn't really make sense to use an sdl3 renderer with sdl2 windowing lol (and i don't think it would even work in the first place)

I'm saying for in the future, also moving the error check before is just standard practice, not really required.

leap0x7b: ah

gradylink (approved):

NateXS:
The 3DS version closes before the Main Menu is on screen (not a crash just exits to homebrew menu).
```
Unzipping Scratch project...
Loading SB3 into memory...
Warning: No unzipped project, trying embedded.
Warning: No sb3 project, trying Main Menu.
Activating main menu...
Main Menu activated.
Successfuly init SDL3!
Cleaned up all text.
```

leap0x7b:
the new commit should fix the issue

NateXS:
It gets to the main menu now, but no input is working currently.

leap0x7b:
should be fixed now (i hope)

NateXS (approved):

PR #517, opened 2025-12-28
Use custom username if setting enabled

roccopm:
I'm unsure if this was an oversight during the getUsername refactor today, but this is how I assume custom usernames are supposed to work. This simply checks if a custom username is enabled in settings and uses it if it is. If not, it falls back to the platform-specific defaults (e.g. Mii name for the Wii), and if there's not one in place, it defaults to "Player".

### Testing
Tested on PC (SDL2), NDS, 3DS and Wii. I have not tested SDL1 or SDL3, though I don't see any reason they would behave any differently.

With a custom username enabled:
Screenshot From 2025-12-28 19-30-27
Screenshot From 2025-12-28 19-35-52
Screenshot From 2025-12-28 19-42-47
image

Without custom username enabled:
Screenshot From 2025-12-28 18-22-26
Screenshot From 2025-12-28 19-36-10
Screenshot From 2025-12-28 19-43-21
image

### DIFF
--- a/source/os.cpp
+++ b/source/os.cpp
@@ -287,6 +287,18 @@ void OS::deInitWifi() {
 }
 
 std::string OS::getUsername() {
+    // use custom username if enabled in settings
+    nlohmann::json json = SettingsManager::getConfigSettings();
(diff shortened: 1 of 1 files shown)


NateXS:
This issue is fixed with 3b225077f23f6f7b60aa1a140859bfb5e9467899.

PR #519, opened 2025-12-29
Optimize/Rewrite(runtime):  to get closer to the behavior of Scratch

Br0tcraft:
The original idea behind this PR was simply to allow value blocks to take multiple frames before returning a value. (things escalated a bit).

There have been a few ongoing issues with execution order, especially when blocks or loops are reseted because something like a costume block is called. To address this more cleanly, every block now returns a BlockResult, including value blocks. This result always contains a default (empty) return value, and also encodes whether the script, sprite, or project should stop, whether the block needs to be executed again in the next frame (for example for glide-like behavior), or whether execution can continue normally.

Previously, a lot of block-specific and execution-specific data lived directly inside the Block struct. This PR starts to separate those concerns. The Block struct itself now only contains block-relevant data such as the linked block function, the next block, and input fields. Everything that varies per “thread” (Scratch calls them threads, though they’re not really threads) is moved into a BlockState, which is stored inside a ScriptThread.

A ScriptThread roughly represents a hat block and the blocks directly below it, and owns the corresponding BlockStates. This allows the same block instance to be used by multiple threads with different state, without needing to duplicate the block itself.

On top of that, I’m experimenting with not loading the entire project zip into RAM at once, but instead extracting images and the project JSON dynamically. This should make it possible to start larger projects. Images could then be managed via an LRU cache and loaded and unloaded on demand.

The project does not compile yet and this is very much a prototype / exploration. Given how frequently the codebase has been changing recently, I’m not entirely sure how far this approach will go, but I wanted to share it early for discussion.
There are still some parts of my code that I want to move, change, etc. (so don't expect too much just yet)


NateXS:
> On top of that, I’m experimenting with not loading the entire project zip into RAM at once, but instead extracting images and the project JSON dynamically. This should make it possible to start larger projects. Images could then be managed via an LRU cache and loaded and unloaded on demand.

I'd personally recommend separating this into a different PR, mainly so the scope of this PR doesn't get *too* big.

source/runtime/runtime.cpp
```diff
@@ -0,0 +1,458 @@
+//runtime.cpp
+#include "runtime.hpp"
+#include "unzip.hpp"
+#include "render.hpp"
+#include "blockExecutor.hpp"
+#include "sprite.hpp"
+struct Broadcasts {
```
gradylink, line 7: Why is this a struct? This should probably be something like a namespace.

source/os.hpp
```diff
@@ -345,3 +358,69 @@ class DirectoryOpenFailed : public FilesystemError {
...
+// ToDo: PLEASE HELP ME
+constexpr uint32_t RAM_USAGE_VALUE = 32; //32 bytes
+constexpr uint32_t RAM_USAGE_BLOCKSTATE = 34; //34 bytes
+constexpr uint32_t RAM_USAGE_EXECUTIONFRAME = 4; //42 bytes ...
+constexpr uint32_t RAM_USAGE_SCRIPTTHREAD = 6;
+constexpr uint32_t RAM_USAGE_BLOCK = 48;
+constexpr uint32_t RAM_USAGE_VARIABLE =  48;
+constexpr uint32_t RAM_USAGE_LIST = 48;
+constexpr uint32_t RAM_USAGE_INPUT = 48;
+constexpr uint32_t RAM_USAGE_SPRITE = 150;
```
gradylink, line 373: This could just use `sizeof`.

source/os.hpp
```diff
@@ -345,3 +358,69 @@ class DirectoryOpenFailed : public FilesystemError {
...
+    uint64_t MAXRAM = 256000000; //~256mb
+#elif defined( __WIIU__)
+    uint64_t MAXRAM = 512000000; //~512mb
+#elif defined( __SWITCH__)
+    uint64_t MAXRAM = 2000000000; //~2gb
+#elif defined(__PS4__)
+    uint64_t MAXRAM = 2000000000; //~2gb
+#else
+    uint64_t MAXRAM = 2000000000; //~2gb
+#endif
```
gradylink, line 425: Should be `constexpr`

PR #520, opened 2025-12-31
Accurate double to string conversion

poipole807:
There's some hacky stuff, but it seems like it works great in games. Probably will have to test more cases though


gradylink (changes requested):
You need to copy the headers to the `include` directory for non-CMake platforms.

gradylink (approved):

PR #538, opened 2026-01-17
Say & Think

roccopm:
New branch for #393 as the history was getting messy after multiple new platforms and codebase refactors. See that PR for screenshots and comment history.

## Tested working platforms
- SDL1 (currently showing a black screen due to a problem in the main branch, but it compiles and no errors appear when loading projects with speech).
- SDL2 (Linux)
- SDL3 (Linux)
- 3DS
- NDS
- Headless (no rendering obviously but made sure it compiles and loads speech projects with no errors)

## Untested platforms
- Haven't tested all SDL2 platforms, but they worked when I tested them on the last PR so no reason they shouldn't still work.
- PS4 (compiles but untested)
- WASM (compiles but untested)

## Unimplemented platforms
- OpenGL (GLFW) is not implemented as it currently appears to be broken on the main branch. Should be straightforward to implement the rendering for this once fixed. There's no dockerfile for this so hoping it's not too big of a deal anyway.

### DIFF
--- a/source/renderers/citro2d/speech_text_c2d.cpp
+++ b/source/renderers/citro2d/speech_text_c2d.cpp
@@ -34,24 +34,14 @@ void SpeechTextObjectC2D::setText(std::string txt) {
 }
 
 std::vector<float> SpeechTextObjectC2D::getSize() {
-    std::vector<std::string> lines;
     std::string text = getText();
     std::istringstream iss(text);
     std::string line;
-    while (std::getline(iss, line)) {
-        lines.push_back(line);
-    }
-    if (!text.empty() && text.back() == '\n') {
-        lines.push_back("");
-    }
-    if (lines.empty()) {
-        return {0.0f, 0.0f};
-    }
 
     float maxWidth = 0.0f;
     float totalHeight = 0.0f;
 
-    for (const auto &line : lines) {
+    while (std::getline(iss, line)) {
         if (!line.empty()) {
             C2D_Text tempText;
             C2D_TextBuf tempBuffer = C2D_TextBufNew(200);
@@ -73,8 +63,8 @@ std::vector<float> SpeechTextObjectC2D::getSize() {
             C2D_TextFontParse(&tempText, *textClass.font, tempBuffer, " ");
             C2D_TextOptimize(&tempText);
 
-            float width, height;
-            C2D_TextGetDimensions(&tempText, scale, scale, &width, &height);
+            float height;
+            C2D_TextGetDimensions(&tempText, scale, scale, nullptr, &height);
             totalHeight += height;
 
             C2D_TextBufDelete(tempBuffer);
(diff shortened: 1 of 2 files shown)


source/renderers/citro2d/speech_text_c2d.cpp
```diff
@@ -0,0 +1,143 @@
...
+        } else {
+            currentLine += c;
+        }
+    }
+    if (!currentLine.empty() || text.back() == '\n') {
+        lines.push_back(currentLine);
+    }
+
+    return lines;
+}
```
gradylink: Using an `std::istringstream` and `std::getline` would be more efficient than using this helper function all together (all the times this is used would be replaced by that).

source/renderers/citro2d/speech_text_c2d.cpp
```diff
@@ -0,0 +1,143 @@
...
+        lines.push_back("");
+    }
+    if (lines.empty()) {
+        return {0.0f, 0.0f};
+    }
+
+    float maxWidth = 0.0f;
+    float totalHeight = 0.0f;
+
+    for (const auto &line : lines) {
```
gradylink: No like what I mean is this bottom loop here should be replaced by the while loop above, also the early return is unnecessary since the loop will be skipped anyway.

source/renderers/citro2d/speech_text_c2d.cpp
```diff
@@ -0,0 +1,143 @@
...
+            C2D_TextBufDelete(tempBuffer);
+        } else {
+            // Empty line still has height
+            C2D_Text tempText;
+            C2D_TextBuf tempBuffer = C2D_TextBufNew(200);
+            C2D_TextFontParse(&tempText, *textClass.font, tempBuffer, " ");
+            C2D_TextOptimize(&tempText);
+
+            float width, height;
+            C2D_TextGetDimensions(&tempText, scale, scale, &width, &height);
```
gradylink: No need to have a `width` variable here, `C2D_TextGetDimensions` allows passing `nullptr`.

gradylink (changes requested):

source/renderers/citro2d/speech_text_c2d.cpp
roccopm: You're right, it was completely redundant. Let me know if you spot any more issues 👍🏼

source/speech_text.cpp
```diff
@@ -0,0 +1,77 @@
...
+std::string SpeechText::wrapText() {
+    if (originalText.empty()) {
+        return originalText;
+    }
+
+    std::string result;
+    std::string currentLine;
+    std::string currentWord;
+
+    for (char c : originalText) {
```
gradylink, line 17: Very very subtle thing but this just has an unnecessary copy for every single character so it should really be `const char &` instead.

roccopm: I believe each character is destroyed before the next iteration, and since a `char` is only 1 byte while a memory address is 4 bytes / 8 bytes (depending on 32bit / 64bit), is it not better to copy it? This seems to be debated in forums and differs from style guide to style guide. Modern compilers will likely produce an equally optimised result either way.

I'm not precious about which is used here but I'd like to hear your thoughts on this before making changes

PR #539, opened 2026-01-20
Clean up speech bubbles when sprites get deleted

roccopm:
Fixes segfault caused by the speech bubble renderer trying to render one relative to a sprite that has been deleted. Just added the speech cleanup to the sprite cleanup. Let me know if I've missed any locations where sprites get deleted.

I also added some needed validation such as 330 max char limit and blank text speech should delete an existing bubble.

### DIFF
--- a/source/speech_manager.cpp
+++ b/source/speech_manager.cpp
@@ -26,8 +26,7 @@ void SpeechManager::showSpeech(Sprite *sprite, const std::string &message, doubl
     clearSpeech(sprite);
 
     if (message.empty()) return;
-    std::string truncatedMessage = message;
-    truncatedMessage = truncatedMessage.substr(0, 330);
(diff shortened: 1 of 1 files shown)


source/speech_manager.cpp
```diff
@@ -25,6 +25,10 @@ void SpeechManager::showSpeech(Sprite *sprite, const std::string &message, doubl
 
     clearSpeech(sprite);
 
+    if (message.empty()) return;
+    std::string truncatedMessage = message;
+    truncatedMessage = truncatedMessage.substr(0, 330);
```
gradylink: Why not just directly assign?

```suggestion
    const std::string truncatedMessage = message.substr(0, 330);
```

NateXS (approved):

PR #546, opened 2026-01-27
[Runtime] Refactor Image code

NateXS:
The goal with this is to remove a lot of the repeated code between the different backends, and *hopefully* make it a bit easier to modify and add on to later down the line. All backends will also now be using stb_image and NanoSVG to load image data.

Progress:
- [x] Base Image class
   - [x] Image loading with stb_image and NanoSVG
   - [x] Ability to render parts of an image using `ImageSubrect`s 
   - [x] Ability to get pixel data from parts of an image
   - [x] Resize bitmap images to half its size (this is because Scratch internally doubles the resolution of bitmap images, so resizing it saves a lot of RAM on low RAM devices) 
   - [x]  Some more, i forgor
- [x] SDL1
- [x] SDL2
- [x] SDL3
- [x] Citro2D
- [x] GL2D
- [x] OpenGL
- [x] Headless

Do all platforms work
- [x] PC
- [x] 3DS
- [x] NDS
- [x] PSP
- [x] Switch
- [x] Wii
- [x] GameCube
- [x] PS Vita (Working in emulator at least)
- [x] Wii U
- [x] PS4
- [x] webOS

This PR also fixes some Sprite rendering issues, mainly with position and rotation being off. This also fixes brightness effect on SDL platforms.

### DIFF
--- a/source/menus/menuObjects.cpp
+++ b/source/menus/menuObjects.cpp
@@ -13,9 +13,9 @@ static float guiScale = 1.0f;
 
 // turn off nineslice for low performance platforms
 #if defined(__NDS__) || defined(__PSP__) || defined(RENDERER_SDL1) /* sdl1 just doesnt support it currently */
-static bool enableNineslice = false;
+constexpr bool enableNineslice = false;
 #else
-static bool enableNineslice = true;
+constexpr bool enableNineslice = true;
 #endif
 
 double MenuObject::getScaleFactor() {
--- a/source/renderers/citro2d/speech_manager_c2d.cpp
+++ b/source/renderers/citro2d/speech_manager_c2d.cpp
@@ -41,9 +41,8 @@ void SpeechManagerC2D::render() {
             int spriteCenterY = static_cast<int>((sprite->yPosition * -scale) + (SCREEN_HEIGHT / 2));
 
             // Calculate actual rendered sprite dimensions
-            double divisionAmount = 1.0;
-            int spriteWidth = static_cast<int>((sprite->spriteWidth * sprite->size / 100.0) / divisionAmount * scale);
-            int spriteHeight = static_cast<int>((sprite->spriteHeight * sprite->size / 100.0) / divisionAmount * scale);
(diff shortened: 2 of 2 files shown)


gradylink:
It would be nice if this PR could add a `getPixel` and `getPixels` (would return a struct with information about the pixel data, like a micro `SDL_Surface` ig along with a raw pixel buffer) for when I finally get around to updating collision.

NateXS:
This PR is pretty much done now, I just need to figure out why PS4 is seemingly crashing.

source/menus/menuObjects.cpp
```diff
@@ -11,6 +11,13 @@ static float guiScale = 1.3f;
 static float guiScale = 1.0f;
 #endif
 
+// turn off nineslice for low performance platforms
+#if defined(__NDS__) || defined(__PSP__) || defined(RENDERER_SDL1) /* sdl1 just doesnt support it currently */
+static bool enableNineslice = false;
+#else
+static bool enableNineslice = true;
+#endif
+
```
gradylink, line 20: These should probably be `constexpr`, the compiler should catch it, but it should probably still be done for clarity reasons.

source/renderers/citro2d/speech_manager_c2d.cpp
```diff
@@ -45,7 +41,7 @@ void SpeechManagerC2D::render() {
             int spriteCenterY = static_cast<int>((sprite->yPosition * -scale) + (SCREEN_HEIGHT / 2));
 
             // Calculate actual rendered sprite dimensions
-            double divisionAmount = sprite->costumes[sprite->currentCostume].isSVG ? 1.0 : 2.0;
+            double divisionAmount = 1.0;
             int spriteWidth = static_cast<int>((sprite->spriteWidth * sprite->size / 100.0) / divisionAmount * scale);
             int spriteHeight = static_cast<int>((sprite->spriteHeight * sprite->size / 100.0) / divisionAmount * scale);
```
gradylink: No reason for the `divisionAmount` variable anymore.

gradylink (approved):

PR #552, opened 2026-02-01
[FIX] DPI scale on init

roccopm:
Also includes a small CMakeLists fix as it couldn't find SDL2_gfx libs and was failing when trying to link.
Tested on MacOS Tahoe 26.2 with SDL2.

Before:
image

After:
image

### DIFF
--- a/source/windowing/sdl2/window.cpp
+++ b/source/windowing/sdl2/window.cpp
@@ -212,7 +212,11 @@ postAccount:
     resize(dw, dh);
 #else
     int dw, dh;
+#ifdef __PS4__
+    SDL_GetWindowSize(window, &dw, &dh);
+#else
     SDL_GetWindowSizeInPixels(window, &dw, &dh);
+#endif
     resize(dw, dh);
 #endif
 


gradylink:
This issue also likely effects the SDL3 windowing, you should update this there as well.

source/windowing/sdl2/window.cpp
```diff
@@ -211,7 +211,9 @@ bool WindowSDL2::init(int width, int height, const std::string &title) {
     SDL_GL_GetDrawableSize(window, &dw, &dh);
     resize(dw, dh);
 #else
-    SDL_GetWindowSize(window, &this->width, &this->height);
+    int dw, dh;
+    SDL_GetWindowSizeInPixels(window, &dw, &dh);
```
gradylink, line 218: PS4's SDL2 port doesn't have this function, I think there's a polyfill you can take from `renderers/sdl2/render.cpp`.

roccopm:
> This issue also likely effects the SDL3 renderer, you should update this there as well.

SDL3 works, I copied the fix from there. Test fails on PS4 for this change though so likely due to lesser implemented SDL2. Will mark as ready when fixed

Issue #396, opened 2025-10-07
[REQUEST] Implement missing "hacked" bloks


yuri-kiss:
### Describe the feature you'd like to propose.

<img width="491" height="422" alt="Image" src="https://github.com/user-attachments/assets/a2c3024e-becf-4cee-b899-be2dd7d62ef7" />

all of the blocks work perfectly fine, the matrix input should just return its binary, it is used in some text engines

Blocks:
- [ ] Stack blocks that do nothing (Align Scene (), Hide All Sprites, etc.)
- [x] Reporter blocks that report `undefined`
- [x] When This Sprite Touches ()
- [x] For Each () in ()
- [x] While ()
- [x] Counter blocks
- [x] All at Once
- [ ] Loud?
- [ ] ~~`data_listindexall`, `data_listindexrandom`~~ (these blocks only exist in the editor, they are converted to their text form before being saved)
- [ ] Instrument blocks
- [x] Pen hue blocks
- [x] Pen shade blocks
- [x] Micro:bit pin state menu
- [ ] Play Note () for () Seconds
- [x] CoreEx Extension

### Pre-submission Checks

- [x] I have checked that the roadmap doesn't already contain this feature.
- [x] I have searched existing issues and confirmed this feature has not been requested before.

KirbySuper008:
(because "for I" wasn't explained) I is a drop-down menu that lists all variables. When the block is run, the variable is set to zero and is incremented each time the repeat loops.

gradylink:
Make sure you're on the latest version because all of these except `when this sprite touches` and `all at once` were added in beta 25.

poipole807:
There's a complete list of "hacked" blocks on the Scratch Wiki: https://en.scratch-wiki.info/wiki/Hidden_Blocks. Most of the unimplemented ones do nothing, but there are four Scratch 2.0 pen blocks that are fully functional and may be worth implementing

Edit: the pen blocks have been implemented

PR #563, opened 2026-02-21
Support undefined

OceanIsEndless:
Hello! This is just a minor change. It adds the hidden no-op reporter blocks mentioned in #396 that report the JavaScript value `undefined`, which has been added as its own type. It's doubtful anyone uses these blocks in this way, but supporting them further improves parity with Scratch. This also fixes the way some values are displayed in data monitors.

- Undefined has been added to Value as its own type. (If focusing only on Scratch blocks, reporting the string "undefined" would technically work, but in the future there may be some way of differentiating undefined from strings that would make this separate type necessary. Some custom extensions would also need this value as a separate type, e.g. `type of ()` in the TW Cast extension reports "undefined" when specifically given `undefined`.)
- The blocks `motion_xscroll`, `motion_yscroll`, and `sensing_userid` all report undefined.
- Scratch will crash when undefined is added to a list monitor, while TurboWarp will display nothing. When you set a variable monitor to undefined on either platform, it will not update until it is set to a value other than undefined. Adding this behavior would be kind of annoying due to the way SE! monitors work, so for now all monitors will show nothing when displaying undefined. While adding this I noticed numbers in variable monitors were not being truncated to six decimal places like in Scratch, so this has been fixed (this is done specifically for variable monitors, as list monitors keep the full number).

I haven't used C++ or contributed code here before, but the changes were relatively small and could be inferred from existing code. There might be a better way to truncate numbers to six decimal places, but besides that, I hope everything looks good!

[Test Project](https://github.com/user-attachments/files/25459984/Test.Project.zip) • [Live Version](https://oceanisendless.github.io/Scratch-Everywhere/?project_url=oceanisendless.github.io/hmm.sb3)

### DIFF
--- a/source/render.hpp
+++ b/source/render.hpp
@@ -195,6 +195,30 @@ class Render {
     }
 
     /**
+     * Gets a variable value as a string for display
+     */
+    static std::string getVariableValueString(Value value) {
+        if (value.isDouble()) {
+            return Math::toString(std::round(value.asDouble() * 1e6) / 1e6); // js Number(value.toFixed(6))
+        } else if (value.isUndefined()) {
+            return ""; // Scratch keeps the original value, leave blank for now
+        } else {
+            return value.asString();
+        }
+    }
+
+    /**
+     * Gets a list value as a string for display
+     */
+    static std::string getListValueString(Value value) {
+        if (value.isUndefined()) {
+            return ""; // Scratch crashes, TurboWarp shows empty string
+        } else {
+            return value.asString();
+        }
+    }
+
+    /**
      * Gets the color for a monitor value background based on opcode
      */
     static ColorRGBA getMonitorValueColor(const std::string &opcode) {
@@ -300,7 +324,7 @@ class Render {
                             drawBox(monitorW - (24 * scale), boxHeight, monitorX + (22 * scale) + (monitorW - (28 * scale)) / 2, monitorY + boxHeight + item_y + (boxHeight / 2), 252, 102, 44);
 
                             std::unique_ptr<TextObject> &itemText = monitorGfx.items[index];
-                            itemText->setText(s.asString());
+                            itemText->setText(getListValueString(s));
                             itemText->setColor(Math::color(255, 255, 255, 255));
                             itemText->setScale(1.0f * (scale / 2.0f));
                             itemText->setCenterAligned(false);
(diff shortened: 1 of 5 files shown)


NateXS (approved):
LGTM!

poipole807:
I would've liked it if this PR wasn't merged so fast, but oh well. There don't seem to be any double or Boolean conversion rules so those would have to be added. Also, undefined reacts with the switch to costume block in weird ways that would have to be tested for maximum accuracy

NateXS:
> I would've liked it if this PR wasn't merged so fast, but oh well. 

my bad 🤒

OceanIsEndless:
> I would've liked it if this PR wasn't merged so fast, but oh well. There don't seem to be any double or Boolean conversion rules so those would have to be added. Also, undefined reacts with the switch to costume block in weird ways that would have to be tested for maximum accuracy

They weren't needed! It already defaults to false and 0 for unrecognized values. Though I don't know about the switch costume one

Edit: You're right, there is something weird with switching costume to undefined. I should have thought how it affects other blocks. I'll look into it

Edit 2: It seems in Scratch, when the input for switch costume to () isn't a number, it tries to convert it to a string like this:

```js
// Strings should be treated as costume names, where possible
const costumeIndex = target.getCostumeIndexByName(requestedCostume.toString());
```

But `undefined` doesn't have a toString method, so it throws an error, and from the console output it just continually runs the switch costume code, throws the same error, and never continues the script... so I suppose for "maximum accuracy" switching the costume to undefined should cause the script to hang and never continue.

I don't know if this undefined behavior should be recreated or not? Using undefined is already unexpected by Scratch, so besides parity, there may not be much benefit in purposely breaking blocks when they are given undefined. Though we've come this far, might as well make it the same...? I'll keep looking to see where else this problem crops up.

Edit 3: Skimming the original implementations of all the blocks, I could only find these ones not accounting for undefined, all for the same reason, and all hanging indefinitely:

- `switch costume to ( v)`
- `switch backdrop to ( v)`
- `switch backdrop to ( v) and wait`

This seems more like a bug than a feature, though bugs have been added for parity in the past (e.g. horizontal tabs being equal to zero). I would report this on the Scratch repo, but then they might as well remove undefined entirely. Is this inconsistency okay for now? As long as other blocks handle values the same way in Scratch and SE!, there shouldn't be any _other_ inaccuracies. ¯\\\_(ツ)\_/¯

yuri-kiss:
it should be noted that some built in extensions also return undefined (see: https://scratch.mit.edu/projects/1253372879/editor/)

image

obviously this is a problem for later and not now

PR #565, opened 2026-02-24
feat(runtime): cache variable/list data & handlers

gradylink:
"Cache" is the best term I could think of.

# Checklist

- [x] Cache block handlers
- [x] Cache variables
	- [x] Cache variables used as field IDs
	- [x] Cache variables used in inputs
- [x] Cache lists
	- [x] Cache lists used as field IDs
	- [ ] ~~Cache lists used in inputs~~ (Decided since like no one ever uses this, we won't handle this case)
- [x] Improve monitors
	- [x] Switch to map
	- [x] List caching
	- [x] Variable caching
- [x] Fix cloning issues
- [x] Heal the cache at runtime (this shouldn't really ever happen, but why not)
- [x] Make optional, a.k.a. evaluate load time (we could also move the link phase into its own thread while loading the project) and RAM impacts.


gradylink:
> [!WARNING]
> Even though I have marked this PR as ready for review, we still need to do the last checklist item due to the fact that this PR increases RAM usage on games like Appel by about 10MB, this is unacceptable for a lot of platforms, so we will need to implement a compile-time option to disable this.

EDIT: this has now been done, you can ignore this.

PR #566, opened 2026-02-24
[SDL2/SDL3] Add Fast Pen project setting

NateXS:
the fast pen option batches all the pen draws to be used in a single `SDL_RenderGeometry()` draw call. 

Fast:
image

Accurate (which is just the old code):
image

The reason why there's two options is because fast pen currently draws everything as rectangles, instead of it being rounded. The accurate option could also use SDL_RenderGeometry but im too lazy to figure out the math right now :grin: :v:

test project: 
[penny.sb3.zip](https://github.com/user-attachments/files/25525907/penny.sb3.zip)


Dogo6647:
oh, didn't see it was already merged when i clicked approve 😭✌️

NateXS:
:skul:

PR #570, opened 2026-02-27
Change the Sans Serif font to the actual font

DevelopCMD:
Self explanitory

### DIFF
--- a/source/image.cpp
+++ b/source/image.cpp
@@ -48,7 +48,7 @@ bool loadFont(const std::string &family, const std::string &path) {
 
 bool Image::Init() {
     if (!loadFont("", "gfx/ingame/fonts/NotoSerif-Regular")) return false;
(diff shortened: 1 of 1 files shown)


DevelopCMD:
this fixes the text being unreadable on some platforms, and also makes it look more accurate to when running the project on vanilla Scratch

gradylink:
Does this also affect the other fonts? Also, you should properly rename the file to reflect this.

DevelopCMD:
but if i rename it, then the code will crash tho

gradylink:
> but if i rename it, then the code will crash tho

Then also rename the file name in the code... I made it extremely simple, it's in `source/image.cpp`.

Dogo6647:
it's as easy as doing Ctrl+F and replacing all references to the old font name to the new one :)

DevelopCMD:
ill rename the font and also change all the instances of it, gimme one sec

DevelopCMD:
that was easy

PR #571, opened 2026-03-01
feat: init svg scaling

gradylink:
# Checklist

- [x] Support All Renderers
	- [x] SDL1
	- [x] SDL2
	- [x] SDL3
	- [x] OpenGL
	- [x] GL2D
	- [x] Citro2D
	- [x] Headless
- [x] Support All Resizable Windowing Solutions
	- [x] GLFW
	- [x] SDL1
	- [x] SDL2
	- [x] SDL3
- [x] Resize on load
- [x] Resize on set size
- [x] Resize on change size
- [x] Scale based on window size
- [x] Resize on window resize.

### DIFF
--- a/source/runtime/unzip.cpp
+++ b/source/runtime/unzip.cpp
@@ -500,8 +500,6 @@ nlohmann::json Unzip::unzipProject(std::istream *file) {
         }
 
         if (keepInRam) {
-            Log::log("keeping project in RAM");
-            filePath.clear();
             Scratch::sb3InRam = true;
 
             // read the file


DevelopCMD:
now that we have everything implemented, it's time to merge this ig

gradylink:
> now that we have everything implemented, it's time to merge this ig

I'm waiting for Nate to review it.

source/render.hpp
```diff
@@ -185,6 +186,34 @@ class Render {
...
+            if (imgFind == Scratch::costumeImages.end()) continue;
+
+            float scale = sprite->size / 100;
+            if (sprite->renderInfo.renderScaleY != 0) scale *= sprite->renderInfo.renderScaleY;
+
+            try {
+                imgFind->second->resizeSVG(scale);
+            } catch (const std::runtime_error e) {
+                Log::logWarning(std::string("Failed to resize SVG: ") + e.what());
+            }
```
NateXS, line 213: Maybe if it fails it could try resizing at a lower scale? because I know devices like the 3DS, NDS and PSP have low texture size limits.

gradylink: The scale is already capped at 5 and then on top of that it'll scale down based on the screen size, so I don't think the textures can get too big in that case.

EDIT: I also already tested a few projects in MelonDS when I was adding GL2D support and didn't notice any issues.

PR #578, opened 2026-03-11
[WIP] feat(playdate): init

gradylink:
Playdate Port!!!

# Checklist

- [x] CMake Stuff
- [x] Initial headless runtime execution
- [x] Support Simulator
- [x] Support Real Hardware
	- [x] Remove exceptions (not supported on real hardware) (#580)
		- [x] Don't use CMakeRC
	- [ ] Get projects loading
		- [ ] Packed (stack overflow, seems to be from a 4KB buffer Miniz allocates)
		- [ ] Unpacked (watchdog timeout)
- [ ] Improve Error Handling
- [ ] Project Loading
	- [x] RomFS Packed
	- [x] RomFS Unpacked
	- [ ] SE! Dir Packed
	- [ ] SE! Dir Unpacked
- [ ] Renderer
	- [ ] Implement `TextObject`
	- [ ] Implement `Image`
		- New image format: 1-Bit
	- [ ] Placeholder sprite rendering
	- [ ] Sprite with costumes
	- [ ] Pen
- [ ] Audio
- [ ] Input
- [ ] Implement Menu (I don't think we'll be able to use the current main menu)
- [ ] Dockerfile/Nightly Action
- [ ] Fix execution speed


ghost:
how is the lever thingy from the playdate gonna work

gradylink:
> how is the lever thingy from the playdate gonna work

I'm thinking mouse y and then mouse down, or some key press indicates whether it's docked?

ghost:
> > how is the lever thingy from the playdate gonna work
> 
> I'm thinking mouse y and then mouse down, or some key press indicates whether it's docked?

maybe, and mouse x?

JoeyL110513:
> > > how is the lever thingy from the playdate gonna work
> > 
> > 
> > I'm thinking mouse y and then mouse down, or some key press indicates whether it's docked?
> 
> maybe, and mouse x?

Maybe the crank can have its own set of variables and blocks or it can control if the player is controlling the mouse or using the d-pad and A+B?

gradylink:
Couldn't figure out the hardware issues for this :( if someone else would like to attempt this it *might* be possble.

KirbySuper008:
insane this was ever a thing 💀
translating color graphics to a 1-bit display is not exactly easy, nor does it typically look good or even discernable

lucas11222:
i may try to do this, any info on why it was closed?

gradylink:
> i may try to do this, any info on why it was closed?

@lucas11222 

The Playdate has an absolutely tiny stack size, I managed to get the app running in the simulator but on real hardware it would stack overflow.

lucas11222:
btw why just not use the sdk directly instead of c++ version of the sdk?

gradylink:
> btw why just not use the sdk directly instead of c++ version of the sdk?

@lucas11222 

Firstly, this entire app is written in C++ and the normal Playdate SDK doesn't have great C++ support. Secondly that doesn't change the stack size at all, we're still using the normal SDK under the hood the only difference is some tweaked build flags.

PR #582, opened 2026-03-14
Rewrite audio stack, and use dr_libs instead for audio files

NishiOwO:


gradylink (approved):

gradylink:
Btw Nate don't merge this yet, tts broke

NateXS (approved):

PR #583, opened 2026-03-14
feat: "Minor" runtime changes

### LINKED ISSUE #500: [BUG] Calling a repeating block while it's already running overrides previous state
Pretty much the title.

Project: https://files.catbox.moe/ou2k34.sb3

Expected output:
```
[PROJECT] hi
[PROJECT] 3
[PROJECT] hi
[PROJECT] 4
[PROJECT] 3
[PROJECT] 4
[PROJECT] 3
[PROJECT] 4
[PROJECT] 4
```

Actual Output:
```
[PROJECT] hi
[PROJECT] 3
[PROJECT] hi
[PROJECT] 4
[PROJECT] 4
[PROJECT] 4
[PROJECT] 4
```

AFAIK, this is one of the last major remaining issues with the runtime.

Originally described here: https://ptb.discord.com/channels/1408875318248345612/1408877980494725120/1449653010555994286

Br0tcraft:
As you already know, I’ve been busy reworking the runtime, and here is the result of that "minor" update. I actually ended up starting from scratch three times and rewrote all the blocks twice because I kept finding new ways to optimize things. It’s been a bit of a journey merging old code back in, so there's a chance I might have overlooked a few details or didn't combine everything perfectly.

The main goal was to implement a "thread" behavior similar to Scratch. It’s not a 1:1 replica yet and has its own C++ specific optimizations and quirks, but it provides a much better foundation. It is now significantly easier and more effective to replicate the original Scratch behavior than it was before. Performance-wise, switching to actual threads instead of constantly copying blocks has already made a huge difference for clone stability and speed, which you might have seen in the video I shared on Discord.

There are still some trade-offs and work to be done. For instance, I haven't re-integrated the variable and list pointer caching we had a few days ago. Even without it, the runtime is often faster in my tests, but it really depends on how much the project mutates variables. More importantly, MyBlocks are currently a bit unstable and tend to crash some projects, which is why the ParityTest and PerformanceTest aren't starting at the moment.

Once those crashes are ironed out, the new architecture should make it very simple to add the TurboWarp "return" extension block.

Edit: I will of course also fix the build errors.


NateXS:
Excited to see this finally get finished!

gradylink:
I wonder if this will fix the weird slow down issue with Appel

EDIT: it does

source/runtime/sprite.hpp
```diff
@@ -23,151 +22,161 @@ struct RenderInfo {
...
+     * If the current block is last in a loop, it will yield for a screen refresh and run the block at the top of the loop.
+     */
+    CONTINUE,
 
-struct ParsedField {
-    std::string value;
-    std::string id;
-};
+    // Continues to the next block, but never waits for the next screen refresh (used for things like loops)
+    CONTINUE_IMIDIATELY,
```
gradylink:
```suggestion
    CONTINUE_IMMEDIATELY,
```

:sob: (obv not the only place this needs to be fixed)

source/runtime/blockExecutor.cpp
```diff
@@ -140,38 +138,163 @@ void BlockExecutor::linkPointers(Sprite *sprite) {
...
+            Scratch::resetInput(currentBlock);
+        }
 
-    while (currentBlock && currentBlock->id != "null") {
-        BlockResult result = executeBlock(*currentBlock, sprite, withoutScreenRefresh, fromRepeat);
+        executionCount++;
+        if (thread.withoutScreenRefresh && var != BlockResult::CONTINUE_IMMEDIATELY) executedBlocks++;
 
-        if (result == BlockResult::RETURN) return;
+        // TODO: re-add (currently makes execution slow, maybe could use a Timer instead?)
```
gradylink: Scratch itself uses a timer iirc so that's definitely what we should do.

Br0tcraft: Wouldn't SE need to query the device time in every frame (possibly even multiple times per frame) if it uses a timer? I thought it would be more performant to simply set a maximum number of blocks per frame (per user-defined block). The maximum can also be set via the project settings

gradylink: > Wouldn't SE need to query the device time in every frame (possibly even multiple times per frame) if it uses a timer? I thought it would be more performant to simply set a maximum number of blocks per frame (per user-defined block). The maximum can also be set via the project settings

Querying device time should be really fast though on pretty much every platform, it *might* have issues on NDS and PSP, but I think it would be worth properly implementing a timer on all the other platforms at least in order to match Scratch's behavior.

gradylink:
There are more memory leaks but I couldn't figure out what was causing them

NateXS (approved):
Apart from any mystery regressions idk about yet, This PR looks good to me!

gradylink (approved):

NateXS:
Merging this in time for Beta 39. Thanks Br0t & everyone who helped!

PR #593, opened 2026-03-28
feat(menu): add "Also try AmpMod"

ampelc:
We're struggling to find users so this is mostly promotion for us. If you don't know, see https://ampmod.codeberg.page

In exchange, we have added a loading message linking to you (in AmpMod 0.5):
image

### DIFF
--- a/gfx/menu/splashText.txt
+++ b/gfx/menu/splashText.txt
@@ -3,6 +3,7 @@ Also try TurboWarp!
 Also try PenguinMod!
 Also try Snail IDE!
 Also try Electramod!
+Also try AmpMod!
 Also try every other Scratch clone!
 Support the Scratch Foundation!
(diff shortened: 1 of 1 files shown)


gradylink:
image

The people have spoken.

The main reasoning behind this decision is that we'd rather not make it appear as though splashes are a method of advertising and merging a PR like this would likely make it appear as though another Scratch mod would be welcome to make a PR to add their mod to our splashes which is not our intended purpose.

ampelc:
> image
> 
> The people have spoken.
> 
> The main reasoning behind this decision is that we'd rather not make it appear as though splashes are a method of advertising and merging a PR like this would likely make it appear as though another Scratch mod would be welcome to make a PR to add their mod to our splashes which is not our intended purpose.

ok

PR #612, opened 2026-04-12
feat(translate): add extension and cache clearing

Br0tcraft:
I'm not entirely satisfied with the current handling of language codes. However, I'm not aware of any other solution that doesn't require parsing another JSON file at runtime or retrieving the language code from a server before translation.

The block works similarly to the Text2Speech block (except that the downloaded file is returned as a value and not played back)

I've also added a "Clear cache" button, as the folder can fill up quickly.
Perhaps an option to enable automatic cache clearing (when the app/project starts) could be added ?

(Tested on Linux)

### DIFF
--- a/source/menus/settingsMenu.cpp
+++ b/source/menus/settingsMenu.cpp
@@ -152,9 +152,8 @@ void SettingsMenu::render() {
     Render::beginFrame(1, 96, 90, 105);
 
     if (ClearCache->isPressed({"a"})) {
-        // clear folder: OS::getScratchFolderLocation() + "cache/"
-        std::filesystem::remove_all(OS::getScratchFolderLocation() + "cache/");
-        std::filesystem::create_directory(OS::getScratchFolderLocation() + "cache/");
+        OS::removeDirectory(OS::getScratchFolderLocation() + "cache/");
+        OS::createDirectory(OS::getScratchFolderLocation() + "cache/");
     }
 
     if (EnableMenuMusic->isPressed({"a"})) {
--- a/source/menus/settingsMenu.cpp
+++ b/source/menus/settingsMenu.cpp
@@ -37,6 +37,10 @@ void SettingsMenu::init() {
     EnableMenuMusic->text->setColor(Math::color(0, 0, 0, 255));
     EnableMenuMusic->text->setScale(0.5);
 
+    ClearCache = new ButtonObject("Clear Cache", "gfx/menu/projectBox.svg", 200, 270, "gfx/menu/Ubuntu-Bold", true);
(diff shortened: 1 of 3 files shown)


gradylink (approved):
lgtm

source/menus/settingsMenu.cpp
```diff
@@ -144,6 +151,12 @@ void SettingsMenu::render() {
     Render::beginFrame(0, 96, 90, 105);
     Render::beginFrame(1, 96, 90, 105);
 
+    if (ClearCache->isPressed({"a"})) {
+        // clear folder: OS::getScratchFolderLocation() + "cache/"
+        std::filesystem::remove_all(OS::getScratchFolderLocation() + "cache/");
+        std::filesystem::create_directory(OS::getScratchFolderLocation() + "cache/");
```
NateXS: should be using `OS::removeDirectory()` and `OS::createDirectory()`

NateXS (changes requested):

NateXS (approved):

OceanIsEndless:
> I'm not entirely satisfied with the current handling of language codes. However, I'm not aware of any other solution that doesn't require parsing another JSON file at runtime or retrieving the language code from a server before translation.

Looked into this in a failed attempt to add the translate extension before the minor runtime changes. The language code stuff is provided by the package [scratch-translate-extension-languages](https://www.npmjs.com/package/scratch-translate-extension-languages), which interfaces with Google Translate. Not sure how feasible it is to simulate that, so for now it would probably be better to keep a copy of the [JSON](https://github.com/OceanIsEndless/stelo/blob/main/dumpedLanguageNames.json). Also tried [optimizing](https://oceanisendless.github.io/stelo/) it to decrease storage when including it in SE!, but you might be able to do better. (A lot of the data is redundant. Managed to slim it to 114 KB while remaining a valid JSON file, could possibly go smaller if not stored in JSON. It contains every language name translated to every other language and maps names to codes, so there's bound to be ways to improve it.)

Br0tcraft:
@OceanIsEndless Thanks for the input! My main concern with a large JSON file is the overhead, it would either need to be parsed repeatedly at runtime or stay resident in memory, which is a significant trade-off on resource-constrained hardware like the 3DS.
I also considered fetching language codes via an API on the fly, but that would effectively double the latency for every translation. Even with the current implementation, the delay is quite noticeable when testing on the 3DS.
I’m planning to revisit the extension once i18n support is officially integrated. This will be necessary anyway to ensure the "language" block returns the correct viewer language instead of defaulting to English. I'll use that opportunity to see if there's a more optimized way to handle the language mapping.

PR #617, opened 2026-04-13
Add Directory Selection Dialog on Desktops

samuelvenable:
(So you won't have to type out the whole path manually)...

WIP, needs a little more testing.

Tiny File Dialogs adds support for both GTK and Qt dialogs on Linux, `*`BSD, and SunOS systems, based on availability.

GTK:
gtk

Qt:
qt

Windows uses standard Windows API, and macOS uses osascript/applescript.


source/process/process.cpp
gradylink, line 1: We do not put external dependencies in the source directory of this project.

gradylink (changes requested):

leap0x7b:
this pr smells vibecoded

source/process/process.cpp
samuelvenable: I'm happy to re-write the tinyfiledialogs code with my own implementation, as I already have my own project which is similar which I could base my code on. As for the process code, I can trim it down to only the code that is actually used/needed for this to be compiled and used. A lot of it I am not even using, as I wasn't initially clear on how you all felt about external deps in your repository. Now that I have that question answered, I can make changes.

Does all of this sound OK to you?

On a different topic, I have the actual dialog as its own executable, because I wanted to make it optional. If the user does not want the dialog to show on startup, maybe we could add a CMAKE flag preventing that portion to build?

samuelvenable: @gradylink `ngs::ps::exe_from_proc_id(ngs::ps::proc_id_from_self())` with the filename removed from the pathname spec can be used to get the executable file's parent directory. It works on Windows, macOS, Linux, FreeBSD, DragonFly BSD, NetBSD, OpenBSD, Solaris, and illumos. Most notibly, I got help from the OpenBSD mailing list to get the OpenBSD solution, because OpenBSD does not have an official function dedicated for the executable path OS-level. It has its failure cases, but it even checks the executable path and sees if the one suspected to be correct matches the correct `ino_t` and `dev_t` and if it doesn't match an empty string is returned with `errno` set to non-zero. This pathname with `"scratch-everywhere/"` appended at the end would probably be better than the working directory with `"scratch-everywhere/"` appended at the end, for the fallback. What do you think?

samuelvenable: > @gradylink `ngs::ps::exe_from_proc_id(ngs::ps::proc_id_from_self())` with the filename removed from the pathname spec can be used to get the executable file's parent directory. It works on Windows, macOS, Linux, FreeBSD, DragonFly BSD, NetBSD, OpenBSD, Solaris, and illumos. Most notibly, I got help from the OpenBSD mailing list to get the OpenBSD solution, because OpenBSD does not have an official function dedicated for the executable path OS-level. It has its failure cases, but it even checks the executable path and sees if the one suspected to be correct matches the correct `ino_t` and `dev_t` and if it doesn't match an empty string is returned with `errno` set to non-zero. This pathname with `"scratch-everywhere/"` appended at the end would probably be better than the working directory with `"scratch-everywhere/"` appended at the end, for the fallback. What do you think?

Maybe I could make this its own pr, since it's unrelated to the changes I'm making in this one?

samuelvenable: > @gradylink `ngs::ps::exe_from_proc_id(ngs::ps::proc_id_from_self())` with the filename removed from the pathname spec can be used to get the executable file's parent directory. It works on Windows, macOS, Linux, FreeBSD, DragonFly BSD, NetBSD, OpenBSD, Solaris, and illumos. Most notibly, I got help from the OpenBSD mailing list to get the OpenBSD solution, because OpenBSD does not have an official function dedicated for the executable path OS-level. It has its failure cases, but it even checks the executable path and sees if the one suspected to be correct matches the correct `ino_t` and `dev_t` and if it doesn't match an empty string is returned with `errno` set to non-zero. This pathname with `"scratch-everywhere/"` appended at the end would probably be better than the working directory with `"scratch-everywhere/"` appended at the end, for the fallback. What do you think?

I finished isolating the relevant code to achieve this into its own source and header: https://github.com/samuelvenable/exepath Let me know if this change interests you.

samuelvenable: > I finished isolating the relevant code to achieve this into its own source and header: https://github.com/samuelvenable/exepath Let me know if this change interests you.

One limitation of the OpenBSD code; there is no way to differentiate between multiple hard links pointing to the same file on disk, even with checking the `ino_t` and `dev_t`, so to prevent erroneous results, it might be a good idea to have the function set errno and return an empty string whenever `st_nlink != 1`.

gradylink: > I'm happy to re-write the tinyfiledialogs code with my own implementation, as I already have my own project which is similar which I could base my code on. As for the process code, I can trim it down to only the code that is actually used/needed for this to be compiled and used. A lot of it I am not even using, as I wasn't initially clear on how you all felt about external deps in your repository. Now that I have that question answered, I can make changes.
> 
> Does all of this sound OK to you?
> 
> On a different topic, I have the actual dialog as its own executable, because I wanted to make it optional. If the user does not want the dialog to show on startup, maybe we could add a CMAKE flag preventing that portion to build?

Why not just use our CMake-based dependency system to pull it from an external repo via CPM?

samuelvenable: I'm not familiar with that. I had no idea cmake had a dependency management system. I'm still relatively new to cmake and am more used to "regular make" and shell scripts for these types of things. I'll do research on how to achieve this. I could tell cmake was doing something with dependencies when I built this project but I wasn't sure how it was doing it. Thanks for the info, I'll correct these issues.

gradylink: > I'm not familiar with that. I had no idea cmake had a dependency management system. I'm still relatively new to cmake and am more used to "regular make" and shell scripts for these types of things. I'll do research on how to achieve this. I could tell cmake was doing something with dependencies when I built this project but I wasn't sure how it was doing it. Thanks for the info, I'll correct these issues.

Well CMake itself doesn't, we built our own, you should look at the `cmake/deps` folder.

samuelvenable: @gradylink I had a look at you dependency management system. It seems to only work with repositories hosted on GitHub, the only official git repository for tinyfiledialogs is on sourceforge. Is there any easy way to make your dependency system work for sourceforge and other non-GitHub git repositories?

Edit:

I don't mind making a GitHub mirror for tinyfiledialogs, but it will quickly get outdated.

gradylink: > @gradylink I had a look at you dependency management system. It seems to only work with repositories hosted on GitHub, the only official git repository for tinyfiledialogs is on sourceforge. Is there any easy way to make your dependency system work for sourceforge and other non-GitHub git repositories?

@samuelvenable

```cmake
CPMAddPackage(
	NAME somepackage
	GIT_REPOSITORY https://example.com/repo.git # URL to any .git repo here
	VERSION 0.0.0
	GIT_TAG release-0.0.0 # can also be inferred from version
	OPTIONS "SOME_OPTION ON" "SOME_OPTION2 OFF" # If needed
)
```

Issue #619, opened 2026-04-13
[REQUEST] AUR Package


Samq64:
### Describe the feature you'd like to propose.

Are there any plans for an [AUR](https://aur.archlinux.org/) package? If not do I have permission to submit one?

### Device(s)

PC

### Pre-submission Checks

- [x] I have checked that the roadmap doesn't already contain this feature.
- [x] I have searched existing issues and confirmed this feature has not been requested before.

Dogo6647:
@Starlii10 is working on this!

Samq64:
By the way I am working on XDG support so packaging is easier.

Edit: I thought `scratch-everywhere` was relative to the executable, not the current working directory.

Starlii10:
Technically I'm not the one responsible for the AUR package, that would be @gradylink. But I did make the initial version of the PKGBUILD and am working on the .desktop file

gradylink:
> Edit: I thought scratch-everywhere was relative to the executable, not the current working directory.

You can also set a custom directory in SE!'s settings when on PC.

Starlii10:
https://aur.archlinux.org/packages/scratch-everywhere

PR #620, opened 2026-04-14
feat(linux): add desktop file

Starlii10:
Required for an AUR package (#619) and generally any Linux package.

Translations should also be added, and I'm waiting for opinions on which icon to use (as of right now, it uses the Itchy icon, since the SE! text icon doesn't have an SVG version).


Starlii10:
I have decided not to include translations for now, but they can be added later.

The unofficial poll I set up in Discord has spoken, and it seems like people prefer the Itchy icon over the SE! text icon, so I'll just keep it like that.

gfx/linux/scratch-everywhere.desktop
```diff
@@ -0,0 +1,14 @@
+[Desktop Entry]
+Name=Scratch Everywhere
```
gradylink:
```suggestion
Name=Scratch Everywhere!
```

gradylink (changes requested):

gradylink (approved):

PR #624, opened 2026-04-18
feat: Interactive CLI Inspector for Real-time Debugging

Br0tcraft:
I finally got around to finishing the Inspector features that have been sitting on the back burner for a while. It’s a debug tool for checking sprite states or variables on the fly without having to clutter the engine with logs. You can now monitor FPS and thread counts, dig into specific sprite data, and even override global or local variables directly through the console while the project is running.

The CLI runs in its own thread to stay responsive, and while I’ve secured the command queue with a mutex, the data inspection itself isn't strictly thread-safe against the main loop. Since this is an optional debug tool that you have to explicitly enable via a flag, I figured those minor race conditions aren't a dealbreaker

I used some AI help to clean up the command parsing and keep the naming conventions consistent, which helped keep the logic straightforward and readable

Let me know what you think!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,8 @@ function(set_preferred_cache_var var_name priority_list valid_options_list doc_s
     set_property(CACHE ${var_name} PROPERTY STRINGS ${valid_options_list})
 endfunction()
 
+cmake_dependent_option(SE_INSPECTOR "Enables the runtime Inspector." ON "SE_PLATFORM STREQUAL pc" OFF)
+
 list(APPEND SE_RENDERER_VALID_OPTIONS "headless")
 if(NOT DEFINED SE_RENDERER_PRIORITY)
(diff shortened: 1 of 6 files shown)


gradylink:
I feel like to avoid confusion, it should still only be enabled if you pass `--inspector` when launching SE! from the command line.

rttyg46305Unj:
ass

Br0tcraft:
@gradylink I added the argument `--inspector` as suggested. Anything else I could improve?

gradylink:
Not that I can think of. I mean it might be nice to make a proper way of registering commands with callbacks, then `help` could be auto generated. But that might be a bit too much work.

Dogo6647:
image
<br>
what is wrong with you

NateXS (approved):
LGTM

rttyg46305Unj:
> image
> what is wrong with you

maybe

PR #639, opened 2026-05-03
feat(translations): init

gradylink:
# Checklist
- [x] Loading of translation keys
- [x] Loading of language specific splashes
- [x] Language option
- [x] Actually add languages other than English (Spanish, Japanese, Korean)
- [x] Better font support
- [x] Add separate font for NDS (new one alone is over NDS RAM limit :skull: )


NateXS (changes requested):
Tested on 3DS and NDS, and neither of them launch

It's weird cus I can't really debug it since it's not technically a crash, it just goes back to the homebrew menu on the 3DS side

Maybe it's related to the font? I also tested on PC and PSP and it worked fine on those platforms.

gradylink:
Very weird.

NateXS (approved):

PR #648, opened 2026-05-05
DECtalk support for TTS

NishiOwO:
image


source/runtime/runtime.cpp
```diff
@@ -82,6 +86,10 @@ std::string Scratch::customUsername;
 std::unordered_map<std::string, std::shared_ptr<Image>> Scratch::costumeImages;
 
 void Scratch::initializeScratchProject() {
+#if defined(ENABLE_DECTALK) && defined(ENABLE_AUDIO)
+    TextToSpeechSafeInit();
```
gradylink, line 103: Shouldn't this be in `main.cpp`'s `initApp` since `Scratch::initializeScratchProject` can be run multiple times.

NishiOwO: true

source/runtime/runtime.cpp
```diff
@@ -82,6 +86,28 @@ std::string Scratch::customUsername;
...
+bool Scratch::initializeRuntime() {
+    Log::deleteLogFile();
+    Render::debugMode = true;
+
+    if (!OS::init()) {
+        return false;
+    }
+    if (!Render::Init()) {
+        return false;
+    }
```
gradylink, line 100: These don't really fit here since they're also needed for the menu.

source/runtime/runtime.cpp
```diff
@@ -82,6 +86,28 @@ std::string Scratch::customUsername;
...
+#ifdef ENABLE_AUDIO
+#ifdef ENABLE_DECTALK
+    TextToSpeechSafeInit();
+#endif
+    if (!SoundPlayer::init()) {
+        Log::logError("Failed to initialize audio.");
+        return false;
+    }
```
gradylink, line 108: Same with `SoundPlayer`.

NishiOwO: image

gradylink (approved):

NateXS (approved):
lgtm

PR #662, opened 2026-05-09
Default to Executable Location for SE Project Directory

samuelvenable:


cmake/deps/__getexecname.cmake
```diff
@@ -0,0 +1,26 @@
+function(_dep_system___getexecname)
+	find_path(__GETEXECNAME_INCLUDE_DIR 
+		NAMES __getexecname/internal.h __getexecname/internal.cpp
+		PATHS /usr/include /usr/include/__getexecname
+	)
+
+	if(__GETEXECNAME_INCLUDE_DIR)
+		add_library(__getexecname STATIC ${__GETEXECNAME_INCLUDE_DIR}/__getexecname/internal.cpp)
+		target_include_directories(__getexecname PUBLIC ${__GETEXECNAME_INCLUDE_DIR})
+	endif()
```
gradylink: ngl we should probably leave the system version empty because source files in the include directory is an antipattern anyway, that'll just force it to use CPM.

You'll also need to add `set(SE_FORCE_SOURCE___getexecname TRUE)` to the top of the file. Other than that, this looks good to me!

gradylink (approved):

NateXS (approved):
lgtm!

samuelvenable:
Thanks guys!

PR #668, opened 2026-05-10
[WIP] N64 port

EGAMatsu:
Unfinished N64 port, doesnt' crash, doesn't display anything.


leap0x7b:
‎source/renderers/n64/ seems to just be a copypaste of ‎source/renderers/opengl/ but with n64 specifics, maybe just modify those instead and have the n64 specifics be on an ifdef?

TheRealMoondude:
knew this was possible

EGAMatsu:
> ‎source/renderers/n64/ seems to just be a copypaste of ‎source/renderers/opengl/ but with n64 specifics, maybe just modify those instead and have the n64 specifics be on an ifdef?

im just trying to get a initial prototype working, there's a non-zero chance I'll dump OpenGL and attempt using libdragon's own 2D API.

MTUnixic:
how's it going

PR #669, opened 2026-05-10
Add DECtalk option (and more)

NishiOwO:
Adds
 - `SE_DECTALK_DEFAULT` option to CMake
 - DECtalk menu
 - change `SE_DECTALK` behavior to only enable DECtalk (and not make it default)
 - adds `ui.pause.dectalk`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,10 @@ cmake_dependent_option(SE_CLOUDVARS "Enable cloud variables." ON "SE_ALLOW_CLOUD
 cmake_dependent_option(SE_DOWNLOAD "Enable the ability to use the internet to download files, such as Text-To-Speech blocks." ON "SE_ALLOW_DOWNLOAD" OFF)
 
 option(SE_MENU "Enables the SE! Main Menu." ON)
-option(SE_DECTALK_DEFAULT "Make DECtalk default TTS." OFF)
+
+if(NOT DEFINED SE_DECTALK_DEFAULT)
+	set(SE_DECTALK_DEFAULT OFF)
+endif()
 if(SE_DECTALK_DEFAULT)
 	option(SE_DECTALK "Build DECtalk for TTS." ON)
 else()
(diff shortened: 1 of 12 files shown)


source/menus/pauseMenu.cpp
```diff
@@ -16,29 +16,47 @@ static std::string getTurboString() {
...
+    return TranslationManager::getTranslation("ui.pause.dectalk") + ": " + TranslationManager::getTranslation(Scratch::useDectalk ? "ui.settings.on" : "ui.settings.off");
+}
+
 void PauseMenu::init() {
+#ifdef ENABLE_DECTALK
+    const int buttons = 4;
+#else
+    const int buttons = 5;
+#endif
+    const int space = (180 - 60) / buttons;
```
gradylink: Very subtle but this should *technically* be `constexpr`.

NishiOwO: fixed

CMakeLists.txt
```diff
@@ -99,8 +99,12 @@ cmake_dependent_option(SE_CLOUDVARS "Enable cloud variables." ON "SE_ALLOW_CLOUD
 cmake_dependent_option(SE_DOWNLOAD "Enable the ability to use the internet to download files, such as Text-To-Speech blocks." ON "SE_ALLOW_DOWNLOAD" OFF)
 
 option(SE_MENU "Enables the SE! Main Menu." ON)
-
-option(SE_DECTALK "Use DECtalk instead for TTS." OFF)
+option(SE_DECTALK_DEFAULT "Make DECtalk default TTS." OFF)
```
gradylink: Maybe instead of this we could do this:

```suggestion
if(NOT DEFINED SE_DECTALK_DEFAULT)
	set(SE_DECTALK_DEFAULT OFF)
endif()
```

Which would allow platforms to set it as well.

NishiOwO: fixed

gradylink (approved):

NateXS (approved):

PR #671, opened 2026-05-10
Music extension

NishiOwO:


rttyg46305Unj:
there should be external soundfont support, as long as its less than 8 mb (it would also only load the corresponding samples for the corresponding instruments and drums i guess)

NishiOwO:
how tho?

rttyg46305Unj:
> how tho?

selecting one from the /project-directory/sf/ or smth

gradylink (approved):
lgtm!

Dogo6647 (approved):

Dogo6647:
who invited bro to review 🥀
image

source/runtime/runtime.cpp
```diff
@@ -103,6 +105,7 @@ bool Scratch::initializeRuntime() {
 #ifdef ENABLE_DECTALK
     TextToSpeechSafeInit();
 #endif
+    Mixer::init();
```
NateXS: I don't like that this is init at the start of the app, since it leaves the ~1MB sf2 buffer in RAM. IMO it would be better to init this when a music block is used, or during parsing if the project uses the extension.

NateXS (changes requested):
I don't know if I'm doing something wrong, but I don't hear any instruments from the extension :(

Tested on macOS and 3DS

source/runtime/runtime.cpp
NishiOwO: this should be fixed now

NateXS (approved):

PR #676, opened 2026-05-13
feat: implement project receiver system

Br0tcraft:
I've been working on a way to send projects to your hardware over the local network so you don't have to mess around with files manually. Right now, it’s working on PC (tested on Linux) and the 3DS. I also tried getting it to run on the Wii U, but I pretty much failed at that for now. I’m going to start fresh in a few days and try to find where the problem is.

To make sharing easy, I wrote a small addon for ScratchBox that could be integrated directly if needed. You just open SE!, go to settings, and hit "receive project." The console then shows a short PIN that you enter in the ScratchBox addon on your PC to link them up. I went with this PIN system because typing in a full IP address in the browser can be a bit of a hassle, and this is just way faster. I'm not a networking expert yet, so there might be a simpler way to do this, but it works.

The code is still a bit of a mess, so I need to clean it up and improve the error handling and logs before it's truly ready. Let me know what you think!


cmake/platforms/3ds.cmake
```diff
@@ -28,13 +29,19 @@ set(SE_BANNERTOOL bannertool CACHE PATH "Path to bannertool executable")
 set(SE_MAKEROM makerom CACHE PATH "Path to makerom executable")
 set(SE_RAM 72 CACHE STRING "The amount of RAM to make available to SE! (MB)")
 
+# Why the hell is this needed now?
+find_program(MKBCFNT mkbcfnt)
+if(NOT MKBCFNT)
+	set(MKBCFNT "/opt/devkitpro/tools/bin/mkbcfnt")
```
gradylink: Should use the `DEVKITPRO` env var, the problem was actually that we weren't doing that before, remove this and replace the lower part with `${DEVKITRPO}/tools/bin/mkbcfont` and it should work.

gfx/translations/en_us.json
```diff
@@ -3,6 +3,9 @@
   "ui.headless.projectPrompt": "Please type what project you want to open.",
   "ui.projects.noProjects": "No Scratch projects found!",
   "ui.projects.path": "Put projects in ",
+  "ui.projects.receive": "Receive Project",
+  "ui.projects.receive.info": "Enter this code on your computer:",
+  "ui.projects.receive.status": "Waiting for data...",
```
gradylink: Shouldn't these be `ui.receive` instead of `ui.projects.receive` since it's its own menu.

source/receiver.cpp
```diff
@@ -0,0 +1,345 @@
...
+        size_t firstSpace = requestLine.find(' ');
+        size_t secondSpace = requestLine.find(' ', firstSpace + 1);
+        if (firstSpace != std::string::npos && secondSpace != std::string::npos) {
+            method = requestLine.substr(0, firstSpace);
+            path = requestLine.substr(firstSpace + 1, secondSpace - firstSpace - 1);
+        }
+    }
+
+    if (method == "OPTIONS" && path == "/send") {
+        const char *preflightResponse =
```
gradylink:
```suggestion
        constexpr const char *preflightResponse =
```

source/receiver.cpp
```diff
@@ -0,0 +1,345 @@
...
+            "Access-Control-Allow-Headers: Content-Type, X-Chunk-Offset, X-Project-Author, X-Project-Title\r\n"
+            "Access-Control-Max-Age: 86400\r\n"
+            "Content-Length: 0\r\n"
+            "\r\n";
+        send(client_fd, preflightResponse, strlen(preflightResponse), 0);
+        close(client_fd);
+        return;
+    }
+    if (method != "POST" || path != "/send") {
+        const char *notFound =
```
gradylink:
```suggestion
        constexpr const char *notFound =
```

source/receiver.cpp
```diff
@@ -0,0 +1,345 @@
...
+    bool hasContentLength = !contentLengthStr.empty();
+
+    std::string folder = OS::getCustomScratchFolderLocation();
+    if (folder.empty()) folder = OS::getScratchFolderLocation();
+    FileSystem::createDirectory(folder.c_str());
+    std::string savePath = folder + fileName;
+
+    std::ofstream outFile(savePath, std::ios::binary);
+    if (!outFile.is_open()) {
+        const char *errResp =
```
gradylink:
```suggestion
        constexpr const char *errResp =
```

source/receiver.cpp
```diff
@@ -0,0 +1,345 @@
...
+        "Access-Control-Allow-Origin: *\r\n"
+        "Content-Length: 2\r\n"
+        "\r\n"
+        "OK";
+    send(client_fd, okResp, strlen(okResp), 0);
+
+    close(client_fd);
+}
+
+std::string ProjectReceiver::getLocalIP() {
```
gradylink: This should probably be an `OS::` function.

CMakeLists.txt
```diff
@@ -358,6 +360,10 @@ if(SE_DOWNLOAD)
 	target_compile_definitions(se-interface INTERFACE ENABLE_DOWNLOAD)
 	se_add_dependency(se-interface libcurl)
 endif()
+if(SE_PROJECTRECEIVER)
+	target_compile_definitions(se-interface INTERFACE ENABLE_PROJECTRECEIVER)
+	se_add_dependency(se-interface libcurl)
```
gradylink: As far as I can tell this is unnecessary.

PR #682, opened 2026-05-15
Update mainMenu.cpp

SplaatSites:
Changed version number to 1.0


gradylink:
We do this on our own when versions actually change.

PR #687, opened 2026-05-18
Fix building when SDL is not in the default include path

ccawley2011:
This is needed when SDL is installed via Homebrew, where SDL is installed in a non-default location, and only the `SDL2` subdirectory is included in the output from `sdl2-config --cflags`. This is only necessary for SDL1 and SDL2 - SDL3 includes the directory containing the `SDL3` subdirectory instead.


gradylink (approved):

NateXS (approved):
lgtm!

PR #688, opened 2026-05-19
Portuguese attachment to end of readme

drestudios:
A portuguese (brazil) version attachment to the end of the readme file


DevelopCMD:
this is completely unnecessary, and its not even formatted properly

gradylink:
I don't really understand why we'd do this, doing it for every language would cause the README to be too long, and I don't want to give some languages special treatment.

Issue #14, opened 2025-07-17
[BUG] Project won't boot


RunTheCoins:
**Project Won't boot**
I was trying to make one of my favorite scratch games work with this runtime, but it seems that it won't even boot :(
it might be an issue with unsupported code but idk

**Scratch Project**
[smos3DS.zip](https://github.com/user-attachments/files/21302739/smos3DS.zip)
here's the project if you want to try and debug it

**Screenshots**
so uh it'll load for a bit then it just sends you back to the homebrew launcher menu

**Device**
Old 3DS, Don't have a way to test New 3DS at this time :(

Chikorita-Lover:
I have the same issue for Beta Release 3 on a New Nintendo 3DS. Again, I don't know if it is something with the project I am using, but after loading for maybe 10 seconds, it returns to the Homebrew Launcher menu. (The loading screen used is the one with squares floating upwards.)

NateXS:
looking into this! @Chikorita-Lover if you could also provide your scratch project as well, that'd be helpful!

Chikorita-Lover:
[Stikman Scratch 3DS (version 1.0.0).zip](https://github.com/user-attachments/files/21302971/Stikman.Scratch.3DS.version.1.0.0.zip)
(Just rename the file extension from .zip to .sb3)

RunTheCoins:
@NateXS did you ever figure out was it wasn't booting?

NateXS:
There's a block in the project that's causing it to close the app, but I haven't figured out which block it is. Still looking into it!

NateXS:
can you see if its working for you in Beta 5?

RunTheCoins:
> can you see if its working for you in Beta 5?

@NateXS I can indeed confirm it boots now

PR #724, opened 2026-07-14
[PC] Add Native File Selection Dialog

samuelvenable:
https://github.com/user-attachments/assets/08eec2bf-cb20-4deb-a988-9587b4d17613

Screenshot_2026-07-14_23-25-04

Screenshot from 2026-07-17 23-27-47


rttyg46305Unj:
looks like its not going great...

samuelvenable:
@rttyg46305Unj idk what you're on about but it's only been two days and this pr is almost ready to be merged.

https://github.com/user-attachments/assets/ef4b78ce-c972-402b-b519-48b67ae289cf

rttyg46305Unj:
> @rttyg46305Unj idk what you're on about but it's only been two days and this pr is almost ready to be merged.
> 
> https://github.com/user-attachments/assets/ef4b78ce-c972-402b-b519-48b67ae289cf
> 

i was talking about the actions not ending successfully

samuelvenable:
> > @rttyg46305Unj idk what you're on about but it's only been two days and this pr is almost ready to be merged.
> > 
> >   
> >     
> >     
> > 
> >     2026-07-15.00-39-49.mov
> 
> i was talking about the actions not ending successfully

I wasn't expecting it to be completed in those commits, anyway. The checks that are failing at this point make less sense because the same exact code is working when I test locally, but I'm sure I'll figure it out before long.


Edit: I'm hoping that if I define the `SDL_VIDEO_DRIVER_COCOA` macro manually on macOS for the SDL-based platforms that should fix the macOS CI, as the undefined 'cocoa' structure member is only really possible when that macro is undefined, according to what I could find in the header. According to gradylink the Windows platform is not broken it just needs the same libraries linked to SE! in the docker file that I already linked to in `CMakeLists.txt`.

Edit2: After checking the CI's cmake configuration logs I found this monstrosity:
> #14 45.51 --   SDL_COCOA                   (Wanted: OFF): OFF

Which would reveal why it's failing to build on that platform with the CI.

Edit3: `___isPlatformVersionAtLeast` is undefined. This is probably because the version of Xcode the CI uses is too old. That should be an easy fix, and just a matter of editing the Xcode version to use in the macOS Docker file. It won't be long before the CI passes now. Not much left to do. This P.R. should only take a day or two more to merge.

samuelvenable:
Alright, macOS CI has been fixed. All that's left now is add the new Windows library dependencies to the Docker file and this should be ready to be merged.

ccawley2011:
It's worth mentioning that for SDL3 builds, there's an API for file dialogs in SDL itself, which should make portability simpler compared to working with native window handles.

https://wiki.libsdl.org/SDL3/CategoryDialog

source/menus/settingsMenu.cpp
```diff
@@ -224,7 +227,13 @@ void SettingsMenu::render() {
     }
 
     if (ChangeFolderPath->isPressed({"a"})) {
-        const std::string newPath = Input::openSoftwareKeyboard(projectsPath.c_str());
+#if defined(_WIN32) || defined(_WIN64) || defined(__APPLE__) || (defined(__linux__) && !defined(__ANDROID__) && !defined(WEBOS)) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__DragonFly__) || (defined(__sun) && defined(__SVR4))
```
ccawley2011: Having a define like `USE_LIBDLGMOD` would allow reducing the amount of times that the list of platforms needs to be repeated.

source/windowing/sdl2/window.cpp
```diff
@@ -29,6 +33,10 @@ bool WindowSDL2::init(int width, int height, const std::string &title) {
     SDL_setenv("VITA_DISABLE_TOUCH_BACK", "1", 1);
 #endif
 
+#if defined(SDL_VIDEO_DRIVER_X11) && ((defined(__linux__) && !defined(__ANDROID__) && !defined(WEBOS)) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__DragonFly__) || (defined(__sun) && defined(__SVR4)))
+    SDL_setenv("SDL_VIDEODRIVER", "x11", 1);
```
ccawley2011: Given that libdlgmod supports kdialog and zenity, it's probably better to check which video driver is in use when getting the window handle rather than enforcing X11 here and breaking Wayland-only systems.

CMakeLists.txt
```diff
@@ -283,11 +283,29 @@ if(SE_EXECUTABLE)
...
 	set_target_properties(scratch-everywhere PROPERTIES OUTPUT_NAME "${SE_OUTPUT_NAME}")
-	if(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
-		target_link_libraries(scratch-everywhere PRIVATE se-interface kvm)
+
+	if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
+		target_link_libraries(scratch-everywhere PRIVATE se-interface comctl32 comdlg32 ole32 oleaut32 uuid shlwapi gdiplus ntdll)
+	elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+		target_link_libraries(scratch-everywhere PRIVATE se-interface "-framework AppKit -framework UniformTypeIdentifiers")
+	elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND NOT WEBOS)
+		target_link_libraries(scratch-everywhere PRIVATE se-interface X11 pthread)
```
ccawley2011: With CMake, you can use [FindX11](https://cmake.org/cmake/help/latest/module/FindX11.html) and [FindThreads](https://cmake.org/cmake/help/latest/module/FindThreads.html) to avoid needing to hard-code platform specific paths and libraries. You can also inherit link libraries from the `libdlgmod` target, which would allow keeping things more self-contained.

samuelvenable:
> It's worth mentioning that for SDL3 builds, there's an API for file dialogs in SDL itself, which should make portability simpler compared to working with native window handles.
> 
> https://wiki.libsdl.org/SDL3/CategoryDialog

I don't necessarily agree with using SDL3's code for this purpose because then that would be more maintenance. Having unified dialog code that works the exact same on all platforms and windowing systems is the proper way to go, IMO. 

Especially since my code is already written and confirmed working in its current state, I don't see much purpose in going back and changing it at this point. 

What I did is more portable, not less. It's the same API on all platforms, minus window handle. Getting the window handle would still not be portable either way, because the approach you suggested is only available to SDL 3.x.

Not counting platform macro checks, the current dialog's window handle code in SDL3 is 2 lines of code,  (I recently omitted the code logic for getting the X11 window handle since that doesn't work on Wayland).
```cpp
#if defined(_WIN32) || defined(_WIN64)
	widget_set_owner(std::to_string((unsigned long long)(void *)SDL_GetPointerProperty(SDL_GetWindowProperties(window), SDL_PROP_WINDOW_WIN32_HWND_POINTER, nullptr)).c_str());
#elif defined(__APPLE__)
	widget_set_owner(std::to_string((unsigned long long)(void *)SDL_GetPointerProperty(SDL_GetWindowProperties(window), SDL_PROP_WINDOW_COCOA_WINDOW_POINTER, nullptr)).c_str());
#endif
```
And the actual directory picker is one line of code in total, a single call to `get_directory_alt(...)` which is implemented and works the same in all windowing systems. It doesn't get much simpler than that. Sorry if I'm being a bit difficult. If you still don't agree I can still apply the changes you wanted, if that means actually getting this merged.

source/menus/settingsMenu.cpp
samuelvenable: Just did it. :)

source/windowing/sdl2/window.cpp
samuelvenable: Completed.

samuelvenable:
@NateXS Ready for review! :)

source/window.hpp
```diff
@@ -1,9 +1,11 @@
 #pragma once
 #include <string>
 
-class Window {
+// 'Window' is a typedef in <X11/Xlib.h>: use 'WindowSE' as class name to avoid conflicts!
```
NateXS, line 4: I feel like having this comment be everywhere the class is referenced instead of only where its defined is unnecessary (same with the Time -> TimeSE comments)

NateXS (changes requested):

NateXS:
code otherwise LGTM and works nicely! (tested on macOS)

source/window.hpp
samuelvenable: Taken care of just now. Sorry for the delay. I was busy with friends the past few days and things have been crazy. When CI passes it should be ready for you to merge. Stand by.

samuelvenable:
@NateXS I don't know if you saw these messages I posted to the discord chat, but I wanted to make you aware of it.

Screenshot 2026-07-20 at 5 27 58 PM

RaphGL:
If you're blocking main thread certain OSes will deem your program as "crashed" and will prompt users to close it. 
This is the reason why SDL3's Dialog API uses a nonblocking poll based API.

samuelvenable:
> If you're blocking main thread certain OSes will deem your program as "crashed" and will prompt users to close it. This is the reason why SDL3's Dialog API uses a nonblocking poll based API.

This literally only happens with SDL. I've tried it with GLFW and raw X11 and that 'not responding' message does not appear at all. This has nothing to do with the OS, and it's also specific to GNOME's WM.

NateXS:
> @NateXS I don't know if you saw these messages I posted to the discord chat, but I wanted to make you aware of it.

Huh yeah this is pretty weird... though I guess it's fine if it's only happening with one WM and the message can be easily ignored. Does this happen when using a specific SDL version or is this with any version?

samuelvenable:
> > @NateXS I don't know if you saw these messages I posted to the discord chat, but I wanted to make you aware of it.
> 
> Huh yeah this is pretty weird... though I guess it's fine if it's only happening with one WM and the message can be easily ignored. Does this happen when using a specific SDL version or is this with any version?

I've verified just now and was able to determine this bug happens with SE!'s GLFW windowing backend as well. A fork of a game engine called ENIGMA, which I also help maintain, does not have this problem with libdlgmod when running under the raw X11 backend, but it does happen with ENIGMA's SDL2 backend. I don't know what to make of this. It's implemented the exact same; the code is the same in both software for libdlgmod and its use.

samuelvenable:
> > If you're blocking main thread certain OSes will deem your program as "crashed" and will prompt users to close it. This is the reason why SDL3's Dialog API uses a nonblocking poll based API.
> 
> This literally only happens with SDL. I've tried it with GLFW and raw X11 and that 'not responding' message does not appear at all. This has nothing to do with the OS, and it's also specific to GNOME's WM.

I must've remembered wrong, this problem happens In GLFW too, but not raw X11.

NateXS (approved):
LGTM!

Issue #643, opened 2026-05-04
[REQUEST] Return "true" for "is Turbowarp?" blocks.


LostShadowGD:
Some games require this and don't start without it. On Scratch this returns false and on Turbowarp it returns true. Could this also return True on SE?

### Device(s)

New 3DS, Old 3DS, DS / DS Lite, DSi, Wii U, Wii, GameCube, Switch, Vita, webOS, PSP, PC, Web, PS4

### Additional context

<img width="521" height="413" alt="Image" src="https://github.com/user-attachments/assets/cfaf413b-630f-42c0-bf5c-26afb7a36eeb" />

### Pre-submission Checks

- [x] I have checked that the roadmap doesn't already contain this feature.
- [x] I have searched existing issues and confirmed this feature has not been requested before.

rttyg46305Unj:
SE! is not as fast as the HTML5 Scratch runtime or Turbowarp.

qwertyad1:
wait wasnt there an "Is SE!?" block?

gradylink:
Yes there is a "is Scratch Everywhere!?" block, maybe we could add a "Spoof TurboWarp" setting?

rttyg46305Unj:
> Yes there is a "is Scratch Everywhere!?" block, maybe we could add a "Spoof TurboWarp" setting?

best way to do it i guess

Spike1377:
Should off by default, especially for low RAM systems.

LostShadowGD:
has this been added? I want to play Seashine Island on my Wii U but it requires Turbowarp

qwertyad1:
> has this been added? I want to play Seashine Island on my Wii U but it requires Turbowarp

just open up the project and change the thing that checks that

PR #731, opened 2026-07-24
Add support for isTurbowarp block

MrNasTechGuy:


BinarryCode:
Could this be a setting as requested here: https://github.com/ScratchEverywhere/ScratchEverywhere/issues/643#issuecomment-4372606900 ?

or would it not be worth it?

PR #732, opened 2026-07-26
[may be badly coded] penguinmod motion blocks

fridayfunkgaming291-cloud:
im quite new to c++ but i did what i can and made like 3 blocks
theres a bunch of stuff missing and also i havent tested this so maybe someone can fix this terrible coding

also the shadow block thing is not included as i have 0 idea on how to use it

i don't know if i'll do the other blocks


rttyg46305Unj:
i dont think its workig...

fridayfunkgaming291-cloud:
that was to be expected i guess

gradylink:
We do not and will not support blocks from Scratch mods or extensions.

rttyg46305Unj:
> We do not and will not support blocks from Scratch mods or extensions.

They could also make a custom lua extension.

PR #733, opened 2026-07-28
feat: add error popup

samuelvenable:
https://github.com/user-attachments/assets/10840fb8-b977-43fc-884a-712182821e86

### DIFF
--- a/source/log.cpp
+++ b/source/log.cpp
@@ -117,7 +117,7 @@ void Log::writeToFile(std::string message) {
 void Log::deleteLogFile() {
     std::string filePath = OS::getScratchFolderLocation() + "/log.txt";
     if (std::remove(filePath.c_str()) != 0) {
-        Log::logError("Failed to delete log file: " + std::string(std::strerror(errno)));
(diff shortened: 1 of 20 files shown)


source/log.cpp
```diff
@@ -54,6 +69,35 @@ void Log::logError(std::string message) {
     writeToFile("<Error> " + message);
 }
 
+// logCritical is a graphical error message
+// Has Abort-button-only when a Fatal Error
+// Adds Ignore button when not Fatal Error:
```
gradylink: If you're going to write comments like this they should be in doxygen format.

samuelvenable: @gradylink you mean like this?
```cpp
/**
 * logCritical is a graphical error message
 * Has Abort-button-only when a Fatal Error
 * Adds Ignore button when not Fatal Error:
 */
```

gradylink (changes requested):
It would be nice for you to also change some important errors to use this new function. Things like failing to load graphical stuff, idk anywhere else u see fit.

gradylink (approved):
lgtm

PR #567, opened 2026-02-25
[WIP] Native Android port

PwLDev:
Adding a little language diversity to SE! with this one. According to the SDL2 docs, only Android 4.4+ is supported, though I'm pretty sure it can go below that with some minor tweaks.
The Java glue code for SDL isn't strictly required to be vendored, it could be copied from the SDL2 sample Android project. Also I'd like to mention that I own a Google Play Developer account, so publishing to the Play Store is totally feasible.

Progress:
- [x] CMake workflow integration
- [x] Rendering
- [x] Project loading
- [x] RomFS
- [x] Fockerdile
- [ ] Optional stuff

By far the easiest SE! port I've done.

PR #734, opened 2026-07-29
feat: check path for dialog helpers

samuelvenable:
Only use the O.G. folder selection method as the fallback when `zenity` or `kdialog` do not exist on the target system, and do not use the fall back when the dialog's "Cancel" button is pressed. Also, I added checks to make sure `cmake` doesn't attempt to use `libdlgmod` on Android, (to respect the near-future support of that platform provided by https://github.com/ScratchEverywhere/ScratchEverywhere/pull/567).


gradylink (approved):
lgtm

samuelvenable:
gracias!

PR #738, opened 2026-07-29
Add Optional Arguments to `Render::Init()`

samuelvenable:


Dogo6647 (commented):
idk but it seems a bit off to me to put the new packaging stuff in their own dedicated directories at the root of the repo, maybe these could be moved to /gfx/platform and integrated with cmake somehow so the process is *also* automated better or spun off to a different repo entirely (e.g. `se-pc-packaging`)?

samuelvenable:
> idk but it seems a bit off to me to put the new packaging stuff in their own dedicated directories at the root of the repo, maybe these could be moved to /gfx/platform and integrated with cmake somehow so the process is _also_ automated better or spun off to a different repo entirely (e.g. `se-pc-packaging`)?

I think it's not good that I conflated/addressed too many issues in one PR. In this PR, I removed the Windows/macOS packaging stuff and left the other changes there. I think PwL was interested in doing some packaging stuff, so I'll let them take care of that separately.

gradylink:
I think it makes more sense for the library build target to instead include the runtime, rendering, and windowing logic, as anyone wanting to use SE! as a library won't want the menu system or `main.cpp` logic (it's really easy to implement on your own if you don't include the menu logic).

source/windowing/glfw/window.cpp
```diff
@@ -29,7 +34,11 @@ bool WindowGLFW::init(int w, int h, const std::string &title) {
     }
 
     glfwWindowHint(GLFW_VISIBLE, GLFW_TRUE);
-    glfwWindowHint(GLFW_RESIZABLE, GLFW_TRUE);
+#if defined(USE_LIBDLGMOD) && defined(SE_USE_LIBRARY_BUILD)
+    glfwWindowHint(GLFW_RESIZABLE, GLFW_FALSE);
+#else
+	glfwWindowHint(GLFW_RESIZABLE, GLFW_TRUE);
+#endif
```
gradylink: Why disable resizing, wouldn't this make more sense as an optional flag like the title?

gradylink (changes requested):
I agree with what was mentioned in Discord where this could be refactored to remove the library-specific features since I have a different (and more versatile) approach to a library target planned.

source/windowing/glfw/window.cpp
samuelvenable: sounds good, doing it as we speak

gfx/translations/de_de.splashes.txt
```diff
@@ -134,4 +135,4 @@ In der Tat
 Ich hab gar nichts gemacht.
 Donaudampfschifffahrtselektrizitätenhauptbetriebswerkbauunterbeamtengesellschaft
 Rinderkennzeichnungsfleischetikettierungsüberwachungsaufgabenübertragungsgesetz
-Hier könnte Ihre Werbung stehen!
\ No newline at end of file
+Hier könnte Ihre Werbung stehen!
```
samuelvenable: @gradylink for the life of me I can't figure out what character has changed on this line. I didn't edit it.

source/runtime/runtime.cpp
```diff
@@ -114,7 +114,7 @@ bool Scratch::initializeRuntime() {
     TextToSpeechSafeInit();
 #endif
     if (!SoundPlayer::init()) {
-        Log::logCritical("Failed to initialize audio.", false);
+        Log::logError("Failed to initialize audio.");
```
samuelvenable: @gradylink also not sure how you feel about this change. I personally find audio not mandatory, as long as continuing execution does not crash the app when initialization fails. A blocking graphical error, even if not fatal, is a bit too distracting imo. If audio doesn't play it's clear something is wrong, as most games have sound.

samuelvenable:
@NateXS gradylink said they don't know what to do with this pr and they would like someone else to handle it. Are you down for giving a review? The only thing remaining I think which needs to be addressed is how we will have these optional arguments such as window width/height/title/resizable modified at runtime. For instance, it could be command line arguments you could optionally pass, or we could even do what SDL does with environment variables as custom hints. What are your thoughts on either approach? I would prefer environment variables personally, just because they would be easier to add.

source/menus/settingsMenu.cpp
```diff
@@ -249,7 +251,12 @@ void SettingsMenu::render() {
 
 #if defined(_WIN32) || defined(_WIN64) || defined(__APPLE__)
 
-        const std::string newPathGui = get_directory_alt(folder_picker_dialog_titlebar_caption, "");
+        std::string newPathGui = get_directory_alt(folder_picker_dialog_titlebar_caption, "");
+
+#if defined(_WIN32) || defined(_WIN64)
+		std::replace(newPathGui.begin(), newPathGui.end(), '\\', '/'); // Normalize path separators
```
samuelvenable: I noticed on Windows there is an excessive trailing slash at the end. For example, if you have the projects path set to `C:\Users\User\Desktop\`, it will show up as `C:\Users\User\Desktop\/`. I noticed if I normalize path separators to be forward slashes on Windows, the double slash at the end is changed to a single slash, which is why I did this. Then, the resulting path will look like `C:/Users/User/Desktop/` instead.

PR #742, opened 2026-08-01
[WIP] iOS/iPadOS/macOS port

PwLDev:
Let's bring native Scratch Everywhere! to the Apple ecosystem, and I have to say this port was easier than Android.
While SE! already works on macOS, this port looks forward to improve experience on macOS, though I don't know how much it would conflict with #738.

- [X] Make it boot up 
- [x] Import/load projects
- [ ] Github Workflow (using macOS workers) ~~Fockerdile (yes, it's possible on Linux)~~
- [ ] Cloudvar support

Lua extensions are not supported.
Sorry old iOS nerds but the lowest version I can target at least on modern macOS Xcode is **iOS 8.0**.
The oldest supported device is the **iPhone 5s**, the first iPhone with ARM64.


Luapree:
anything before the android port /j

Dogo6647:
> anything before the android port /j

trur

PwLDev:
Some updates about this port:
- You can now import files directly to the app's folder in iOS 11+ directly within the files app
- Associated the .sb3 extension with the app, the copy implementation is still pending
se_ios_files
se_ios

PR #746, opened 2026-08-08
[WIP] Minor Runtime Changes

Br0tcraft:
So, I originally just wanted to see if I could squeeze a bit more performance out of the runtime, but things got slightly out of hand and turned into a pretty big overhaul. Since quite a few new files came along with this, I also took the opportunity to clean up and restructure the runtime code a bit while I was at it.

Under the hood, I replaced the old execution flow with a brand new bytecode system and introduced a few macros to keep things clean. Instead of just defining its execution code, a block now also dictates how it should be loaded. I also built constant folding directly into the parser definition. If the parser sees that a block returns a constant value (as long as its inputs are constant), it evaluates it right away during parsing, which means things like if-else blocks can completely optimize away unused branches before runtime even starts.

To keep the CPU cache happy, I switched data layouts over to a Structure of Arrays (SoA) setup. Sprites are now managed in an EntityManager struct as components spread across vectors and accessed through a SpriteID. Accessing properties through IDs adds a slight indirection compared to our old direct pointers, but cloning sprites is much lighter now since it doesn't need to copy the TargetDefinition anymore, and renderOrder is reduced to just a single vector of IDs. Alongside that, I updated the Value class to use a String Pool, so equality checks now quickly compare unique string IDs before falling back to a case-insensitive string comparison.

A neat side effect of eliminating raw pointers and references during parsing is that we could potentially serialize all of this straight into a binary file in the scratch-everywhere folder. Replacing or reworking the unpack-project feature with something like that could make project startup almost instant down the road.

In terms of current progress, I haven't translated that many blocks over yet, but a basic script like "When green flag clicked -> move 10 steps" already compiles down cleanly in the terminal to bytecode like 0 [PUSH POS INT], 10 [NUMBER FROM ABOVE INSTRUCTION], 5 [motion_move_steps], and 8 [RETURN]. On the rendering side, things might still be a little broken. I got it building and running on Linux via SDL2, but the build configs and setups for other platforms will definitely need some love.

Since so many blocks are still missing, I haven't been able to run any real performance benchmarks yet, so this is mostly a "hope it runs faster" situation for now. The theory behind it makes sense, but considering this is my first time writing a bytecode interpreter, I'm not entirely sure if my naming or architectural choices are optimal. I'm honestly just happy that it actually runs in practice.

There is still quite a bit of work left, so any help would be super appreciated! If anyone wants to jump in to help adapt the renderers for other platforms or help translate and integrate all the remaining blocks, that would be amazing. Let me know what you think of the approach!


source/menus/settingsMenu.cpp
```diff
@@ -19,6 +18,11 @@
 #include <filesystem.hpp>
 #include <log.hpp>
 
+// um idk but somehow PATH_MAX wasn't defined so I defined it myself
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
```
gradylink: You should remove this and instead merge in main, I fixed this in https://github.com/ScratchEverywhere/ScratchEverywhere/commit/93149e72044a8177954cbecdcb489423a89b2d81

DevelopCMD:
why did the minor changes break the whole thing

DevelopCMD:
should we close this

Dogo6647:
> why did the minor changes break the whole thing

> should we close this

you clearly don't understand how things work around here do you

PR #747, opened 2026-08-09
Add More Critical Errors.

samuelvenable:


gradylink (changes requested):
What about the other renderers, they all have similar error logs.

samuelvenable:
> What about the other renderers, they all have similar error logs.

this pr wasn't finished yet, I was planning on doing that

samuelvenable:
> What about the other renderers, they all have similar error logs.

@gradylink you said all the other renderers have this problem. I checked and that is not true. sdl2 and sdl3 are literally the only two that apply here. This should be ready for being merged now.

PR #750, opened 2026-08-13
Update Dockerfile.3ds

iCraft7773:
I fixed the Dockerfile, so custom project wouldn't give an error. This Dockerfile works with both compiling SE! and any custom Project. Hope my Dockerfile helps!

### DIFF
--- a/docker/Dockerfile.3ds
+++ b/docker/Dockerfile.3ds
@@ -1,25 +1,20 @@
 FROM debian:bookworm AS makerom_installer
 WORKDIR /makerom
-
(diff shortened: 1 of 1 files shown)


iCraft7773:
Also I have 2 sreenshots here. 
The first one is the build with the normal Dockerfile and the build command from the SE! Documentation https://scratcheverywhere.github.io/docker. As you can see, it gives an error. In the second screenshot I used my own Dockerfile (which is in this Pull Request) and the normal command from SE! Documentation. My Dockerfile fixes the issue.
Dockerfile1 normal build command


Dockerfile2 normal build command

NateXS:
Thanks!

PR #755, opened 2026-08-24
[WIP] Ubuntu Touch "port"!

Dogo6647:
[Information about Ubuntu Touch](https://docs.ubports.com/en/latest/)
[Ubuntu Touch developer documentation](https://docs.ubports.com/en/latest/appdev/index.html)

I say "port" in quotation marks because all this is is an extension of the PC port, targetting an older SDL2 version for native rendering support on UT 16.04.5, applying commodity platform-specific changes (via the new special `__UBUNTUTOUCH__` compile flag), and built from the Ubuntu Touch 16.04.5 SDK, which is essentially a Docker image that can compile a program into many different architectures contains the UT sysroot, and is entirely configured by the `clickable.yaml` file at the root of the repo. Additionally, it also packages the application into UT's standard package file format (click).

To build this port, simply install [Clickable](https://clickable-ut.dev/en/latest/install.html), then run 
```
clickable build
```
in the project's root directory. Currently, it will error out near the end of the process as packaging isn't yet set up correctly, but it *will* generate a binary at `/build/ut/{arch}/scratch-ut`.

- [X] Builds a binary
- [ ] Boots to menu
- [ ] Assets/packaging (partially done)
- [ ] App installs in the actual OS
- [ ] Loads/runs projects
  - Menu
    - [ ] Packed
    - [ ] Unpacked
  - RomFS
    - [ ] Packed
    - [ ] Unpacked
- [ ] UT-specific runtime fixes/changes
  - [ ] os.cpp/platform properties (partially done)
  - [ ] Rendering+window stuff (partially done)
  - Other stuff maybe
- [ ] Dockerfile (umm...)
- [ ] Cloud variables (I don't see why they wouldn't work but we'll see)


Dogo6647:
Port update: Stuck, currently looking for a C++17-compatible way to cross-compile this for armhf and arm64 on the Ubuntu 16.04 environment the SDK provides; no cross-compiling gcc8 packages are provided in the PPA repo I'm using.

Dogo6647:
Managed to fully switch to LLVM/Clang 8 and `libc++8`, currently figuring out some (seemingly minor) `std::filesystem` implementation issues.

